PR #11, opened 2022-12-03
Add depends-on hint and newline to bld_ef3__course_char__combined_wid…

jayckaiser:
…e.sql; fix is_empty_model() macro to run on execute and reference get_single_value() macro within its namespace.

This PR fixes four issues with the current implementation of 'bld_ef3__course_char__combined_wide':
1. The macro is_model_empty() always returns False during compile, so an if-execute conditional is required.
2. The macro is_model_empty() raises an error that its helper macro get_single_value() is missing, unless its package namespace is referenced.
3. The ref to 'xwalk_course_level_characteristics' is within a jinja-conditional, so a depends_on SQL hint is required during compile.
4. If 'xwalk_course_level_characteristics' is empty, the compiled SQL combines the last column with the 'FROM' of the select, so the jinja end-if needs its terminal whitespace.

### DIFF
--- a/macros/is_empty_model.sql
+++ b/macros/is_empty_model.sql
@@ -5,7 +5,7 @@
     select count(*) from {{ ref(model_name) }}
   {%- endset -%}
 
-  {%- set nrow = get_single_value(sql_statement) -%}
+  {%- set nrow = edu_wh.get_single_value(sql_statement) -%}
 
   {{ return(nrow == 0) }}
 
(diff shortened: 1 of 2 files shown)


ejoranlienea (changes requested):
> The macro is_model_empty() always returns False during compile, so an if-execute conditional is required.

This appears untrue. Let's remove the `if execute` condition and the hint.

jayckaiser:
It appears you are correct. I believe I was getting false positives for the execute-conditional and the hint. I completed a full DBT run after their removal.

ejoranlienea (approved):

PR #23, opened 2023-01-18
Feature/spec ed program services

rlittle08:
**UPDATE** we decided to make the fct for program services generic to program type (special ed, homeless, etc.). We'll need follow up(s) to add those other types to the table, but the name and structure can stay as is.

For handling [special education program services](https://schema.ed-fi.org/datahandbook-v320/index.html#/SpecialEducationProgramService2765)

- create build model that lists the program service names for a stu-prog association, into an array column
- bring that array column into special ed program assoc table
- create fact table for long stu-special ed program-program services (uses the stg table that makes them long) . extract extensions that are specific to the program service, which is a thing in Boston

Dependent on: https://github.com/edanalytics/edu_edfi_source/pull/10

Questions:
- should null program services be included in the build table?
- this might have to be repeated for other program types e.g. student homeless program services. is there a more efficient way to handle all program services, instead of proliferating into multiple build and fact tables?
- how to guide users to correctly aggregate from the fact table?


rlittle08:
remove comment about NULLs in bld

models/core_warehouse/fct_student_special_education_program_association.sql
```diff
@@ -46,13 +50,21 @@ formatted as (
...
     from stg_stu_spec_ed
     inner join dim_student
         on stg_stu_spec_ed.k_student = dim_student.k_student
     inner join dim_program
         on stg_stu_spec_ed.k_program = dim_program.k_program
+    -- left join because not all special ed programs include services (and they're optional in EdFi)
+    left join bld_program_services 
+        on stg_stu_spec_ed.k_student = bld_program_services.k_student
+        and stg_stu_spec_ed.k_program = bld_program_services.k_program
+        and equal_null(stg_stu_spec_ed.k_lea, bld_program_services.k_lea)
```
rlittle08, line 65: drop these joins

ejoranlienea (approved):

PR #24, opened 2023-01-19
Feature/chronic absenteeism update

ejoranlienea:
This branch introduces two features and two potentially breaking changes to `fct_student_daily_attendance`:

1. **Make all indicators like `is_absent` floats instead of booleans**
  - This allows for fractional attendance codes (like half days), and makes it easier to sum indicators (without coercing them first)
  - Also makes it easier to introduce fully proportional daily attendance (computed from section attendance) without changing the table signature
  - **Migration:** `xwalk_attendance_events` must switch from `True/False` values in `is_absent` to numbers between 0 and 1. Existing T/F values can be swapped for `1.0` / `0.0`
  - **Migration:** Metrics should not coerce these indicators to integer to allow for fractional attendance

2. **Include daily attendance values for students after their enrollment ends.**
  - All indicators (absent, present, enrolled) are set to 0 after or between enrollments, so sums on these values still work as expected
  - This allows daily or over-time calculations of Chronic Absenteeism metrics, since these include students who were but are no longer enrolled. On days after or between enrollments, a student's absence rate, cumulative counts, and metric values stay as they were on their last enrolled day. If a student re-enrolls at the same school, counts continue accumulating after that day.
  - **Migration:** Average Daily Attendance style metrics should either filter out rows where `is_enrolled = 0` to avoid counting students not actually eligible, or make sure they are doing `sum(is_enrolled)` rather than `count(*)` to avoid inflating denominators.

Also introduces a configuration test to ensure all attendance codes are specified between 0 and 1, as setting a value like `50` instead of `0.5` would break the logic.

### DIFF
--- a/models/core_warehouse/fct_student_daily_attendance.sql
+++ b/models/core_warehouse/fct_student_daily_attendance.sql
@@ -37,8 +37,6 @@ school_max_submitted as (
 attendance_calendar as (
     -- a dataset of all possible days on which school attendance could be recorded
     select 
-        --todo: detecting the maximum attendance submitted thus far?
-
         dim_calendar_date.k_school,
         dim_calendar_date.k_school_calendar,
         dim_calendar_date.k_calendar_date,
@@ -63,15 +61,13 @@ stu_enr_att_cal as (
         enr.entry_date,
         attendance_calendar.k_calendar_date,
         attendance_calendar.calendar_date,
-        sum(1) over(
-            partition by enr.k_student, enr.k_school
-        ) as total_days_enrolled
+        enr.exit_withdraw_date
     from fct_student_school_assoc as enr
     join attendance_calendar
         on enr.k_school_calendar = attendance_calendar.k_school_calendar
-    where (attendance_calendar.calendar_date <= enr.exit_withdraw_date
-          or enr.exit_withdraw_date is null)
-    and attendance_calendar.calendar_date >= enr.entry_date
+    -- keep days from enrollment to current-date/end of year to assist with rolling
+    -- absenteeism metrics forward post-enrollment
+    where attendance_calendar.calendar_date >= enr.entry_date
 ),
 fill_positive_attendance as (
     select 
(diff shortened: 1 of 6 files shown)


models/core_warehouse/fct_student_daily_attendance.sql
```diff
@@ -85,15 +81,31 @@ fill_positive_attendance as (
...
+        end as attendance_event_category,
+        coalesce(
+            case 
+                when is_enrolled = 1 then fct_student_school_att.is_absent
+                else 0.0
+            end, 0.0) as is_absent,
+        -- invert is_absent: 1->0, 0->1, 0.25->0.75
+        1.0 - coalesce(
+            case 
+                when is_enrolled = 1 then fct_student_school_att.is_absent
```
rlittle08, line 106: should this be "not fct_student_school_att.is_absent"?

models/core_warehouse/fct_student_daily_attendance.sql
```diff
@@ -138,14 +150,15 @@ cumulatives as (
...
+            partition by k_student, k_school) as total_days_enrolled,
+        sum(is_absent) over(
             partition by k_student, k_school 
             order by calendar_date) as cumulative_days_absent,
-        sum(is_present::integer) over(
+        sum(is_present) over(
             partition by k_student, k_school 
             order by calendar_date) as cumulative_days_attended,
-        sum(1) over(
+        sum(is_enrolled) over(
```
rlittle08, line 161: there's now no difference between total_days_enrolled and cumulative_days_enrolled, so do we need both columns?

rlittle08 (changes requested):

models/core_warehouse/fct_student_daily_attendance.sql
ejoranlienea: No -- since `is_absent` is no longer boolean, we have to do this `1 - is_absent` math.
Basically: if `is_absent` = 1, then 1-1 = 0
`is_absent` = 0, then 1-0 = 1
`is_absent` = 0.5, then 1-0.5 = 0.5

models/core_warehouse/fct_student_daily_attendance.sql
ejoranlienea: They're still different -- cumulative counts up throughout the year, while total is always the final total. Either can be useful for inclusion business rules.

ejoranlienea:
A couple new changes:
- Exclude rows from the enrollment table where `exit_withdraw_date` is either before the first day of school, or before the students reported `entry_date` -- consider these invalid enrollment records.
- Add an audit table that captures all dropped enrollment records for inspection if necessary.

PR #25, opened 2023-01-19
Feature/domain disabling

jayckaiser:
This PR adds model property yaml files to each subdirectory of the project. Each model and test is given a tag representing a given domain/program, and non-core models are enabled/disabled based on the presence of 'src' variables in the project environment.

The only exception to this rule are the models in core_warehouse. Many already had separate property yaml files, so I've added new ones where missing.


jayckaiser:
I have rebased and added tags and enabled configs for the new student program models added by Rob in PR 23.

ejoranlienea (approved):

PR #26, opened 2023-01-27
Feature/parents

jalvord1:
Included in this feature:
- build models to cast parent emails and phones wide, with the resulting column names being determined by xwalks (Note: the need for xwalks results from special characters in the original names - It would be easy to switch all special characters to an empty string but that would be an issue for `/` or anything analogous, which we've seen already)
- parent dimension that merges on the above build tables as well as a full mailing address ('home' option chosen when available)
- fact table with student parent associations (Note the logic used to determine 'current' associations)
- two custom tests:
    1. one that checks that a student only has a single primary contact (severity: warn)
    2. one that checks whether a given student_xyr/parent association is linked only in a single `school_year` (severity: error) - this may seem weird because we link the most recent student/parent association to all `k_student`s for a `k_student_xyear` but the `school_year` will remain the same, which is why there should still only be a single `school_year` for a `k_student_xyear`/`k_parent`

This feature requires that `feature/full_address` in the source package is also merged.

### DIFF
--- a/models/build/edfi_3/parents/bld_ef3__parent_wide_emails.sql
+++ b/models/build/edfi_3/parents/bld_ef3__parent_wide_emails.sql
@@ -7,8 +7,8 @@ parent_email_types as (
 emails_wide as (
   select 
     k_parent,
-    -- todo: do I need api year in here?
-    -- or do I dedupe to keep latest?
+    tenant_code,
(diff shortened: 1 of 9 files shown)


jalvord1:
@ejoranlienea tested the changes with both empty and non-empty xwalks, good to go now

ejoranlienea (approved):

PR #35, opened 2023-02-13
Feature/optional program associations

jayckaiser:
This branch standardizes and renames all program association build models such that they are structured identically. This branch and [feature/standardize_program_associations in edu_edfi_source](https://github.com/edanalytics/edu_edfi_source/pull/20) are co-dependent and must be merged together.

This branch changes the column names of program start and end dates in particular, so this will break existing Stadium implementations that define 'edu:special_ed:start_date_column' as "spec_ed_program_begin_date" (as well as the end-date column).

Both branches will require thorough code review to verify they are feature-complete. Both have been tested successfully against Stadium South Carolina.


models/core_warehouse/dim_student.sql
```diff
@@ -63,9 +79,53 @@ formatted as (
         stu_grade.entry_grade_level as grade_level,
         stu_races.race_ethnicity,
 
+        -- student programs
         {% if var('src:program:special_ed:enabled', True) %}
-            coalesce(stu_annual_spec_ed.is_special_education_annual, false) as is_special_education_annual,
-            coalesce(stu_is_spec_ed.is_special_education_active, false) as is_special_education_active,
+            {% if 'annual' in var('edu:special_ed:agg_types') %}
```
rlittle08, line 84: could this be a for loop over program types?

models/core_warehouse/fct_student_special_education_program_association.sql
```diff
@@ -1,14 +1,15 @@
 {{
   config(
     post_hook=[
-        "alter table {{ this }} add primary key (k_student, k_student_xyear, k_program, spec_ed_program_begin_date)",
+        "alter table {{ this }} add primary key (k_student, k_student_xyear, k_program, k_school, program_enroll_begin_date)",
```
rlittle08, line 4: k_school isn't needed in pkey, it's part of k_program

macros/any_var.sql
```diff
@@ -0,0 +1,34 @@
...
+        {{ any_var([
+              'src:program:special_ed:enabled',
+              'src:program:homeless:enabled',
+              'src:program:language_instruction:enabled',
+              'src:program:title_i:enabled'
+            ], default=True
+        ) }}
+    ```
+#}
+{% macro any_var(var_list, default=True) %}
```
rlittle08, line 18: where is this macro used?

rlittle08 (changes requested):

models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
```diff
@@ -0,0 +1,60 @@
...
+        max(homeless_primary_nighttime_residence) as homeless_primary_nighttime_residence
+
+    from stage
+    group by 1, 2
+),
+
+xyear_agged as (
+    select
+        k_student_xyear,
+        boolor_agg(is_homeless_annual) as is_homeless_ever
```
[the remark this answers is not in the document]
jayckaiser: I agree with this, and I'll look for a solution that does not require this function explicitly.

macros/any_var.sql
jayckaiser: This macro is not used currently. The hope was to use it in the properties config yaml for `fct_student_programs`, but the enabled field cannot be populated by a macro. I still wonder if there's a workaround that will use the macro, so I haven't removed it yet.

models/core_warehouse/dim_student.sql
jayckaiser: Yes this can, and that's a good idea that I'll implement.

models/core_warehouse/fct_student_special_education_program_association.sql
jayckaiser: Understood, and will be removed!

susanxiong:
@jayckaiser  Jeffco is going to be using this! Do you need any support for testing it out? Could @sleblanc23 and I pull it and test it out in jeffco dev (i.e. is it ready for that)?

models/build/edfi_3/students/bld_ef3__student_program__special_education.sql
```diff
@@ -0,0 +1,60 @@
+{# customizable: certain program names may be excluded and not counted as language instruction #}
```
susanxiong, line 1: typo: should read "customizable: certain program names may be excluded and not counted as special education"

jayckaiser: Good catch!

models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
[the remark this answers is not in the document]
jayckaiser: That looks good, @rlittle08. However, I wonder if it can be simplified. Could a `max()` be used cross-platform instead?
https://stackoverflow.com/a/58620918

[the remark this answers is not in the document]
rlittle08: as in, don't use the boolor_agg function for snowflake, use max for all?

also do you mean max without casting to int?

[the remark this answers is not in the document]
jayckaiser: I wonder whether we could just use `max()` for all platforms, ignoring `boolor_agg()` altogether. There may be performance-differences between the two methods, but given the data-size, I don't expect these to be limiting.

I also don't believe we will need a cast to int. Maybe this won't be the case in other platforms, but min/max logic should be identical whether working with booleans or numerics.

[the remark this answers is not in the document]
rlittle08: you can't `max()` booleans in Postgres :(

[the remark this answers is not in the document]
jayckaiser: Ah darn. So there's a decision to be made here, depending on the depth to which we want to build cross-platform variability in our code.

Option 1: `max(is_homeless_annual::integer)::boolean`
Option 2: `{{ boolor_agg('is_homeless_annual') }}`

I don't think it likely that a `boolor_agg()` function will be used in too many places, so I'd consider whether it's worth keeping at all in favor for generic and built-in methods, even if they're slightly less performant or unsightly.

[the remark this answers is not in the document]
jayckaiser: I've chatted with Erik. When we get into cross-database support, we can revisit how this method is implemented. For now though, in Snowflake, this max will work fine.

ejoranlienea (approved):

PR #37, opened 2023-03-07
feature/student program associations

sleblanc23:
Currently a draft - opening this PR to start getting feedback.

This feature will add is_active/is_annual/is_ever columns for a configurable set of programs to dim_student using data from the `studentProgramAssociations` endpoint. It requires a new xwalk added to the project template: xwalk_student_programs.

At this point, the build model and modifications to dim_student are tested in Jeffco and working as expected.


ejoranlienea:
I'm guessing we'll need to split `program_type` into two fields: one that matches the name in the database, and another that's safe to use as a column name; consider the case where the original Program Type is very long, or has spaces or special characters. The xwalk would then be a mapping between values like `Special Education Program` and `spec_ed`.

Another thing to consider is whether the program-type to indicator mapping could ever be many-to-one. Not sure if we want to allow for this, but in principle using the `in` operator instead of `=` would allow for both `IEP` and `504` to be mapped to `spec_ed`.

sleblanc23:
`program_type` was a bad naming decision on my part - I think it should actually be `indicator_name`. I'm not using the program_type field from the studentProgramAssociations data, only the one defined in the cross walk as the intended column name. I added a screenshot of my current version of the Jeffco crosswalk.

The `select distinct` in the `program_types_query` definition should take care of the many-to-one scenario but I will test this.

image

models/build/edfi_3/students/_edfi_3__students.yml
```diff
@@ -48,6 +48,10 @@ models:
     config:
       tags: ['title_i']
       enabled: "{{ var('src:program:title_i:enabled', True) }}"
+  - name: bld_ef3_student_programs
+    config:
+      tags: ['programs']
```
jayckaiser, line 53: I actually think this model should be tagged with 'base' instead of 'programs', since student characteristics and indicators are also populated by default, based on the xwalks.

sleblanc23:
@jayckaiser I added the ability to choose which aggregates are included in the build model (and therefore in `dim_student`) for each indicator by specifying it in the xwalk. I went this route instead of defining `edu:programs:agg_types` variable(s) because I thought it would be a cleaner way to allow for different choices for each indicator and wouldn't require changes to `accordion_columns`. If we end up keeping this approach, I think a test would be helpful to make sure the xwalk has uniform agg types specified for each indicator if there's a many-to-one situation.
program xwalk

sleblanc23:
The only thing is it requires `xwalk_student_programs` to run - that would go in the project template?

ejoranlienea (approved):

PR #38, opened 2023-03-09
added dim_education_service_center

sleblanc23:
Added dim model and yml file for education service centers, modeled off `dim_lea`. Tested and working in Boston.

One small concern - I called the key `k_service_center` in the staging model, but that doesn't totally follow the convention of matching the dim model name. We could go with `dim_service_center`, `k_education_service_center`, or convert to using `dim_esc` and `k_esc` if we want better consistency.

### DIFF
--- a/models/core_warehouse/dim_education_service_center.sql
+++ b/models/core_warehouse/dim_education_service_center.sql
@@ -1,0 +1,45 @@
+{{
+  config(
+    post_hook=[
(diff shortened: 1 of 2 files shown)


rlittle08:
looks good, but maybe add `website` & change to `k_education_center` once new staging branch is in

rlittle08 (approved):

PR #40, opened 2023-03-21
fix for implementations w/ extensions - use union relations & handle extensions

rlittle08:
The `stacked` CTE in `fct_student_program_service` was breaking on Boston, because some relations being stacked have extension columns, but some don't. 
https://github.com/edanalytics/edu_wh/blob/bbcf95b5bb2bf51101822330fc1e4b22817bcf11/models/core_warehouse/fct_student_program_service.sql#L44-L58

### This fix branch does:
- change the stack to a union relations, so that cols are filled in with NULL if they don't exist in one of the programs relations (e.g. extension column "service_duration" does not exist in the homeless program relation)
https://github.com/edanalytics/edu_wh/blob/830c24fc2654bf5aaf3748645ef42bad9c015e85/models/core_warehouse/fct_student_program_service.sql#L44-L51

**EDIT 4/3/23 - code here is now simpler, thanks to an update to `extract_extension` that allows for a list of model names**

- ~~loop over the extension column names from `var("extensions")` to select them in the final query
https://github.com/edanalytics/edu_wh/blob/830c24fc2654bf5aaf3748645ef42bad9c015e85/models/core_warehouse/fct_student_program_service.sql#L53-L59~~
- ~~use "unique" so that if the same column name exists as an extension in multiple relations, they are stacked into the same column in the final query
https://github.com/edanalytics/edu_wh/blob/830c24fc2654bf5aaf3748645ef42bad9c015e85/models/core_warehouse/fct_student_program_service.sql#L73-L78~~
- get each name of the enabled `stage_program_relations` https://github.com/edanalytics/edu_wh/blob/10041e2d01a1707bec6f77bcd6b9885cc77eda32/models/core_warehouse/fct_student_program_service.sql#L53-L57, and call `extract_extension` on that list of relation names 
https://github.com/edanalytics/edu_wh/blob/10041e2d01a1707bec6f77bcd6b9885cc77eda32/models/core_warehouse/fct_student_program_service.sql#L71

### Dependent on:
- https://github.com/edanalytics/edu_edfi_source/pull/24

### Tests done:
- stadium_boston, main branch, extensions correctly pulled out
- stadium_boston, main branch edited to set `extensions: null`, code correctly compiled with no extensions pulled out

### Questions:
- Is there a way to pull out the extensions w/ fewer lines of code?
- ~~This exposes a shortcoming of the `extract_extensions` macro. Is there a way to improve that macro for a case like this, where extensions from multiple models are pulled together? Or create a new macro?~~ **done, see https://github.com/edanalytics/edu_edfi_source/pull/24**


models/core_warehouse/fct_student_program_service.sql
```diff
@@ -42,26 +42,44 @@ dim_program as (
...
+    stacked.program_enroll_begin_date,
+    stacked.program_service,
+    stacked.primary_indicator,
+    stacked.v_providers,
+    stacked.service_begin_date,
+    stacked.service_end_date
+
+    {# -- Select the extension columns from each relation (use unique so there's one column per extension, even if it exists in multiple relations) #}
+    {%- for extension in extensions|unique -%}
+        {% if loop.first %}, {% endif %}
```
jayckaiser, line 75: @rlittle08 I'd recommend moving `stacked.v_providers` after the extension loop in order to simplify the comma syntax. You'd be able to default the comma-suffix and remove any call to the `loop` object attributes.

jayckaiser:
I like this setup, and I agree that altering `edu_edfi_source.extract_extension()` to accept multiple relations would be a fitting change. We can overload the `model_name` parameter to accept either a single string, or a list of strings, and include the `unique` logic in the macro directly.

Let me know your thoughts and if you'd like me to prototype this fix.

rlittle08:
@jayckaiser oh I like that idea for `extract_extension`. Maybe we could pair program it on Thursday?

jayckaiser (approved):
Abstracting extension logic into the `extract_extension()` macro in `edu_edfi_source` was the right call.

We should consider overloading `extract_extension()` to accept a list of relations as well as name strings (as relations are the default argument type passed into most `dbt_utils` macros), but this would involve a minor refactor of multiple models.

For now, however, the solution that maps relations to names works well and is not difficult for a newcomer to the code to understand.

ejoranlienea (approved):

PR #46, opened 2023-03-31
add post hook to create k_network foreign keys using xwalk

rlittle08:
Implementations may have arbitrarily many types of networks. these are configured in `xwalk_network_association_types`, and come through dim_school as `k_network__{{region_type}}`. This branch adds a post hook that loops over those network types and creates a foreign key to `dim_network`.

tests:
- boston with current set up (1 network type in xwalk, 1 in data)
- boston with imaginary set up (2 network types in xwalk, 1 in data) -- this is why I needed to add a `;` in between looped statements. **is that dangerous?**
- tested with the `set network_types` to top of script to avoid duplicating code, but this doesn't work. I believe post hooks are compiled separately, so it needs to be defined in both places

### DIFF
--- a/models/core_warehouse/dim_school.sql
+++ b/models/core_warehouse/dim_school.sql
@@ -1,7 +1,18 @@
+-- depends_on: {{ ref('dim_network') }}
 {{
   config(
     post_hook=[
         "alter table {{ this }} add primary key (k_school)",
+
+        "{% set network_types = dbt_utils.get_column_values(
+                           table=ref('xwalk_network_association_types'),
+                           column='network_type',
+                           where=\"association_type = 'school'\",
+                           order_by='network_type')
+        %}
+        {% for network_type in network_types %} 
+            alter table {{ this }} add constraint fk_{{this.name}}_{{network_type}}_network foreign key (k_network__{{network_type}}) references {{ ref('dim_network') }} (k_network)
+         {% endfor %}"
     ]
   )
 }}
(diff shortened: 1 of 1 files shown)


models/core_warehouse/dim_school.sql
```diff
@@ -38,7 +50,7 @@ formatted as (
         %}
         {%- if network_types is not none and network_types | length -%}
           {%- for network_type in network_types -%}
-            bld_network_associations.k_network__{{network_type}},
+            bld_network_associations.k_network__{{network_type}}::varchar(32) as k_network__{{network_type}},
```
ejoranlienea, line 53: what's the purpose of this `varchar(32)`?

It should already be of type `string`, no? Do we need to coerce it again, and if we do, do we need to put on a length limit?

rlittle08: if not casted to `varchar(32)`, it defaults to `varchar(16777216)`  and the foreign key fails because: `Primary key and foreign key data type does not match`. Every surrogate key dbt creates is `varchar(32)`

rlittle08: it looks like `k_network` goes from length 32 to 16777216 in the build model pivot https://github.com/edanalytics/edu_wh/blob/8e9dfc7d93951d693ec2b3472799d1acefa1f68e/models/build/edfi_3/bld_ef3__wide_school_network_assoc.sql#L53

ejoranlienea (approved):

PR #53, opened 2023-05-02
Feature/accordian columns coalesce

sleblanc23:
This feature came from looking at the Jeffco student programs indicators in `dim_student`, which are sparsely populated with true/false for students involved in at least one program but are null for most students. We would prefer to have those null values shown as false, to avoid confusion with what null vs. false means. These columns are added to `dim_student` using the `accordion_columns` macro.

I added an optional parameter called `coalesce_value` to the `accordion_columns` macro. When this value is provided, a coalesce function is added with the returned column as the first parameter and the `coalesce_value` as the second.

I added a `coalesce_value` to the student programs and student characteristics columns, since both only contain boolean columns. Student indicators are not necessarily all booleans, so I left that one unchanged. I tested this in Jeffco and it is working as intended.

### DIFF
--- a/macros/accordion_columns.sql
+++ b/macros/accordion_columns.sql
@@ -6,8 +6,11 @@ be included in a table.
     in the table
   - source_alias: the alias of the table within the sql expression. If none (default)
     it will reuse the source_table name.
+  - coalesce_value: if this value is specified, a coalesce function will be added with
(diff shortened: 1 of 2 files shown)


susanxiong (approved):
lgtm!

PR #55, opened 2023-05-08
Bugfix/student assessment other results

sleblanc23:
The inner join in these two models was excluding any records that did not have any of the 'standard' results (scale score, SEM, or performance level). For example, CMAS includes records with no scores and a reason that the scores were invalidated. Changing these to left joins allows all records to be included in the fact models, even if they have no standard results or  `xwalk_assessment_scores` and `xwalk_objective_assessment_scores` are empty.


jalvord1 (approved):

susanxiong:
@ejoranlienea anything blocking this from getting merged? Noah/Jeffco are eager to see CMAS and MAP in prod.

PR #62, opened 2023-06-14
restrict district, school ownership by tenant

ejoranlienea:
In Texas, each district-level ODS individually defines _all_ LEAs and Schools in the entire state. This leads to an undesired outcome, where all districts and schools have a separate row in `dim_lea` and `dim_school` for every tenant in the warehouse, and neither table is unique by `lea_id` or `school_id` alone.

This new build model and join in `dim_lea` uses `studentSchoolAssociation` to determine which tenants _actually_ contain which districts, and filtering out the rest. Since `dim_school` depends on a join to `dim_lea`, this restriction also cascades there.

A weakness here is that if district A ever pushed an enrollment of a student at district B, this build model would then contain two LEAs for that tenant, leading to duplication again. This seems unlikely, but may be possible.

### DIFF
--- a/models/build/edfi_3/bld_ef3__tenant_lea_ownership.sql
+++ b/models/build/edfi_3/bld_ef3__tenant_lea_ownership.sql
@@ -1,0 +1,10 @@
+-- in multi-tenant systems, it is possible for each tenant to individually 
+-- define all the districts/schools in a state. We want to avoid having 
+-- each tenant individually define each district and school, and only
+-- keep the definition from the tenant that actually has ownership
+-- since the relationship between tenant and district is not codified anywhere,
+-- we observe it from student enrollments.
+select distinct ssa.tenant_code, lea_id
+from {{ ref('stg_ef3__student_school_associations') }} ssa 
+join {{ ref('stg_ef3__schools') }} sch 
+    on ssa.k_school = sch.k_school
--- a/models/core_warehouse/dim_lea.sql
+++ b/models/core_warehouse/dim_lea.sql
@@ -9,7 +9,9 @@
 with stg_lea as (
     select * from {{ ref('stg_ef3__local_education_agencies') }}
 ),
-
+tenant_lea_ownership as (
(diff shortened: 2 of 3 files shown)


models/build/edfi_3/bld_ef3__tenant_lea_ownership.sql
rlittle08, line 1: Are we ok with allowing the possibility of an lea_id being owned by multiple tenants? That would only happen if it were found in multiple tenants' stu-school data. I can't decide if that's a feature vs. bug

models/core_warehouse/dim_lea.sql
```diff
@@ -46,6 +48,9 @@ formatted as (
     from stg_lea
     left join choose_address 
         on stg_lea.k_lea = choose_address.k_lea
+    join tenant_lea_ownership
```
rlittle08, line 51: should we make this optional & default to include all schools?

models/build/edfi_3/bld_ef3__tenant_lea_ownership.sql
```diff
@@ -0,0 +1,10 @@
+-- in multi-tenant systems, it is possible for each tenant to individually 
+-- define all the districts/schools in a state. We want to avoid having 
+-- each tenant individually define each district and school, and only
+-- keep the definition from the tenant that actually has ownership
+-- since the relationship between tenant and district is not codified anywhere,
+-- we observe it from student enrollments.
+select distinct ssa.tenant_code, lea_id
+from {{ ref('stg_ef3__student_school_associations') }} ssa
```
rlittle08, line 8: if ssa is empty (not yet populated), this will effectively drop all schools from the tenant. Can we detect that and default back to including all schools?

rlittle08 (commented):
This feature is needed for Rally (cc: @jalvord1).

I like this solution, but I added a couple qs about edge cases & whether it should be optional/default.

PR #64, opened 2023-07-06
immutable student demographics

ejoranlienea:
This PR creates a category of demographic variables that the warehouse will consider 'immutable', in the sense that new data is treated as corrections to the record rather than incremental updates.

Since the student dimension is annualized, we keep a record of what a student's demographics were as of each school year. However this can have negative consequences in some cases: if a student's name changes across years we may want to avoid having their prior name appear in reporting.

This branch designates a subset of student demographics as immutable, collects the most recently observed version of these records, and applies them across all history.

An alternative approach would be to make this set of immutable characteristics user-configurable, or otherwise present current and historic versions of records in a more nuanced way.


rlittle08:
@ejoranlienea I added the on/off switch in both models, tested it on boston with no config, true and false values. I decided not to rename `bld_ef3__immutable_stu_demos` for now, but we could do that if you want

PR #65, opened 2023-07-12
Feature/assessment normalizations

rlittle08:
### Overview
Assessment score results in the ODS are often not in the format needed to present in downstream dashboards, or in analytical queries. This branch adds the ability to create "normalized" columns in `fct_student_assessment`, `fct_student_objective_assessment`, where each implementation can customize how they map these values. For example, you may want to map a "Percentile" result on a 1-100 scale to be a "performance_level" on a 1-5 scale. Or, you may want to add ordered integer values e.g. "Low -> 1; Medium -> 2; High -> 3".

### Description of Changes
- In `bld_ef3__student_assessment_long_results` and `bld_ef3__student_objective assessment_long_results`, add `normalized_score_result` to model, to allow for normalization of results
- In `fct_student_assessment` and `fct_student_objective_assessment`, create new columns `normalize_{score_name}` wherever they have been configured. e.g. if you add rows to `xwalk_assessment_score_values.csv` for `performance_level`, a new column `normalized_performance_level` will be created with normalized score results

### Dependent on:
These xwalks added to implementation repo:
- `xwallk_assessment_score_values`
- `xwalk_assessment_score_value_thresholds`
- `xwalk_objective_assessment_score_values`
- `xwalk_objective_assessment_score_value_thresholds`

Example PR: https://github.com/edanalytics/stadium_txdemo/pull/9

### Questions:
- Is "normalized" the right wording for these kind of customizations? I don't want to confuse "display values" with "re-scaled values", but there is overlap there
- Is it right to overload a general "normalized_" column with various use cases, when some may need to be integers vs. characters, etc.?
- Generally, are `edu_wh` models the correct place for this kind of normalization?
- When should re-mapping of values live in student warehouse tables, vs. in dimensions or xwalks?

### TODOs:
- Work on a larger "assessments engine" feature that separates normalized info from assessment-specific info & efficiently serves those data to downstream purposes
- In the future, we may need to allow these xwalks to join separately on subject, grade level, etc.

### DIFF
--- a/models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
+++ b/models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
@@ -4,6 +4,12 @@ with score_results as (
 xwalk_scores as (
     select * from {{ ref('xwalk_assessment_scores') }}
 ),
+xwalk_score_values as (
+    select * from {{ ref('xwalk_assessment_score_values') }}
+),
+xwalk_score_value_thresholds as (
+    select * from {{ ref('xwalk_assessment_score_value_thresholds') }}
+),
 performance_levels as (
     select
         tenant_code,
@@ -34,16 +40,35 @@ dedupe_results as (
 ),
 merged_xwalk as (
     select
-        tenant_code,
-        api_year,
-        k_student_assessment,
-        score_name as original_score_name,
-        coalesce(normalized_score_name, 'other') as normalized_score_name,
-        score_result
+        dedupe_results.tenant_code,
+        dedupe_results.api_year,
+        dedupe_results.k_student_assessment,
+        dedupe_results.score_name as original_score_name,
+        coalesce(xwalk_scores.normalized_score_name, 'other') as normalized_score_name,
+        -- todo should we rename this to original_?
+        dedupe_results.score_result,
+        coalesce(xwalk_score_value_thresholds.normalized_score_result::varchar,
+                 xwalk_score_values.normalized_score_result::varchar,
+                 score_result::varchar
+                ) as normalized_score_result
     from dedupe_results
     left join xwalk_scores
         on dedupe_results.assessment_identifier = xwalk_scores.assessment_identifier
         and dedupe_results.namespace = xwalk_scores.namespace
         and dedupe_results.score_name = xwalk_scores.original_score_name
+    left join xwalk_score_values
+        on dedupe_results.assessment_identifier = xwalk_score_values.assessment_identifier
+        and dedupe_results.namespace = xwalk_score_values.namespace
+        and xwalk_scores.normalized_score_name = xwalk_score_values.normalized_score_name
+        and dedupe_results.score_result = xwalk_score_values.original_score_result
+    left join xwalk_score_value_thresholds
+        on dedupe_results.assessment_identifier = xwalk_score_value_thresholds.assessment_identifier
+        and dedupe_results.namespace = xwalk_score_value_thresholds.namespace
+        and xwalk_scores.normalized_score_name = xwalk_score_value_thresholds.normalized_score_name
+        -- todo check these comparators -- what if there's a value between the upper and next lower? eg value is 20.4 and the cutoffs are 20 and 21
+        -- todo review my use of try_to_numeric here -- the idea is to allow numeric values to merge, otherwise don't merge without error
+        and try_to_numeric(dedupe_results.score_result) >= xwalk_score_value_thresholds.lower_bound
+        and try_to_numeric(dedupe_results.score_result) <= xwalk_score_value_thresholds.upper_bound
+
 )
 select * from merged_xwalk
--- a/models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
+++ b/models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
@@ -45,7 +45,6 @@ merged_xwalk as (
         dedupe_results.k_student_assessment,
         dedupe_results.score_name as original_score_name,
         coalesce(xwalk_scores.normalized_score_name, 'other') as normalized_score_name,
-        -- todo should we rename this to original_?
         dedupe_results.score_result,
         coalesce(xwalk_score_value_thresholds.normalized_score_result::varchar,
                  xwalk_score_values.normalized_score_result::varchar,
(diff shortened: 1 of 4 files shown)


models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
```diff
@@ -34,16 +40,35 @@ dedupe_results as (
...
-        score_result
+        dedupe_results.tenant_code,
+        dedupe_results.api_year,
+        dedupe_results.k_student_assessment,
+        dedupe_results.score_name as original_score_name,
+        coalesce(xwalk_scores.normalized_score_name, 'other') as normalized_score_name,
+        dedupe_results.score_result,
+        coalesce(xwalk_score_value_thresholds.normalized_score_result::varchar,
+                 xwalk_score_values.normalized_score_result::varchar,
+                 score_result::varchar
```
jalvord1, line 51: We talked here whether or not the original score result should be defaulted to if there is no normalization happening for the score and leaned toward yes for the case when normalization is not necessary. What this _could_ mean is that a score that should be normalized but isn't yet included in either normalization xwalk will make it's way into this column in an ugly format that might not match what is necessary for reporting, but in order for a column to be included here, it _must_ be added to the `xwalk_assessment_scores` column in the first place, so there is at least a manual step that needs to happen anyway. Someone might not know that this normalized column exists and the values in the normalized column should be an integer if it's a performance level (as a random example), but I think we can communicate this out and it avoids having to map values to themselves.

models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
```diff
@@ -34,16 +40,35 @@ dedupe_results as (
...
+        and dedupe_results.score_result = xwalk_score_values.original_score_result
+    left join xwalk_score_value_thresholds
+        on dedupe_results.assessment_identifier = xwalk_score_value_thresholds.assessment_identifier
+        and dedupe_results.namespace = xwalk_score_value_thresholds.namespace
+        and xwalk_scores.normalized_score_name = xwalk_score_value_thresholds.normalized_score_name
+        -- todo check these comparators -- what if there's a value between the upper and next lower? eg value is 20.4 and the cutoffs are 20 and 21
+        -- todo review my use of try_to_numeric here -- the idea is to allow numeric values to merge, otherwise don't merge without error
+        and try_to_numeric(dedupe_results.score_result) >= xwalk_score_value_thresholds.lower_bound
+        and try_to_numeric(dedupe_results.score_result) <= xwalk_score_value_thresholds.upper_bound
+        -- todo in future, may need to include subject & grade level in this join (with options to join across subjects)
```
jalvord1, line 71: we will definitely run into this at some point but can start without it - especially considering there will be additional assessment normalization features anyway

models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
```diff
@@ -34,16 +40,35 @@ dedupe_results as (
...
+        and dedupe_results.namespace = xwalk_score_values.namespace
+        and xwalk_scores.normalized_score_name = xwalk_score_values.normalized_score_name
+        and dedupe_results.score_result = xwalk_score_values.original_score_result
+    left join xwalk_score_value_thresholds
+        on dedupe_results.assessment_identifier = xwalk_score_value_thresholds.assessment_identifier
+        and dedupe_results.namespace = xwalk_score_value_thresholds.namespace
+        and xwalk_scores.normalized_score_name = xwalk_score_value_thresholds.normalized_score_name
+        -- todo check these comparators -- what if there's a value between the upper and next lower? eg value is 20.4 and the cutoffs are 20 and 21
+        -- todo review my use of try_to_numeric here -- the idea is to allow numeric values to merge, otherwise don't merge without error
+        and try_to_numeric(dedupe_results.score_result) >= xwalk_score_value_thresholds.lower_bound
```
jalvord1, line 69: this will default to int since no scale argument is given, I think that's fine but maybe we consider allowing for decimals (so try_to_decimal)? I assume you could still write out the values in the xwalk as integers

models/core_warehouse/fct_student_assessment.sql
```diff
@@ -50,6 +52,18 @@ student_assessments_wide as (
             else_value='NULL',
             agg='max',
             quote_identifiers=False
+        ) }},
+        {#- find distinct score names that are in one of the normalize_result xwalks (distinct scores to add normalized_ column for) -#}
+        {% set normalized_names_values = dbt_utils.get_column_values(ref('xwalk_assessment_score_values'), 'normalized_score_name') or [] %}
+        {% set normalized_names_thresholds = dbt_utils.get_column_values(ref('xwalk_assessment_score_value_thresholds'), 'normalized_score_name') or [] %}
+        {{ dbt_utils.pivot(
+            'normalized_score_name',
+            (normalized_names_values + normalized_names_thresholds) | unique,
```
jalvord1, line 61: idea here is that we only want normalized versions of scores that are included in either xwalk (because scores like `scale_score` and `sem` will rarely be normalized in this way, so would be overkill in my opinion)

jalvord1 (commented):
As an overall comment, these are minimal enough changes that I think it's fine to add to `edu_wh` despite the fact that we will continue to add more assessment reporting/normalization features down the line. I think there are no really strong reasons to not add normalized score columns to the fact tables, even if we end up doing more normalization downstream in the future, since score normalization is typically the first ask

models/build/edfi_3/assessments/bld_ef3__student_assessments_long_results.sql
rlittle08: that's a good point, maybe we should be explicit about the data type of this column -- i'm still unsure about this Q I put in the PR "Is it right to overload a general "normalized_" column with various use cases, when some may need to be integers vs. characters, etc.?"

jalvord1: Yeah that's a good q. I think in a lot of cases though the point of a column like this is to normalize values to a similar set of values across all assessments in the table. I don't necessarily think that's always true but my guess is this column would be used for a single particular downstream purpose - like a BI user might use a normalized column where all PLs are integers when creating charts for proper ordering. But again, maybe there is another use case I'm not considering where this could have serious negative effects

rlittle08:
outdated, closing

PR #72, opened 2023-09-06
Bring custom columns from discipline xwalks

susanxiong:
## Description & motivation
Use accordion_columns macro to allow us to bring in custom columns from xwalk tables. Will bring in all columns in the xwalk (outside of the standard model columns). Will do nothing if there are no additional columns outside of the standard, so this is a non-breaking change.

Issue: https://educationanalytics.monday.com/boards/3556827529/pulses/5089328641

- `models/core_warehouse/fct_student_discipline_actions.sql`: add accordion_columns macro
- `models/core_warehouse/fct_student_discipline_incident_behaviors.sql`: add accordion_columns macro

## New models created:
none

## Tests and QC done:
- Tested in Jeffco both with and without custom columns in the xwalks.
- Tested in Boston with no custom columns and confirmed dbt still ran successfully.

## PR Merge Priority:

Low

### DIFF
--- a/models/core_warehouse/fct_student_discipline_actions.sql
+++ b/models/core_warehouse/fct_student_discipline_actions.sql
@@ -95,6 +95,11 @@ join_descriptor_interpretation as (
         xwalk_discipline_actions.is_iss,
         xwalk_discipline_actions.is_exp,
         xwalk_discipline_actions.is_minor,
+
+        -- bring in any additional custom columns from the xwalk table
+        {{ accordion_columns('xwalk_discipline_actions',
+            exclude_columns=['discipline_action','severity_order','is_oss','is_iss','is_exp','is_minor']) }}
+
         xwalk_discipline_actions.severity_order,
         -- for a specific discipline event (which can include multiple disciplines)
         -- flag the most severe discipline
--- a/models/core_warehouse/fct_student_discipline_actions.sql
+++ b/models/core_warehouse/fct_student_discipline_actions.sql
@@ -96,7 +96,7 @@ join_descriptor_interpretation as (
         xwalk_discipline_actions.is_exp,
         xwalk_discipline_actions.is_minor,
 
-        -- bring in any additional custom columns from the xwalk table
+        -- bring in any additional custom columns from xwalk, does nothing if there are no extra columns
         {{ accordion_columns('xwalk_discipline_actions',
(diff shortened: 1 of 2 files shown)


jalvord1:
@susanxiong the changes here look good, we'll probably want to combine this with the [other discipline branch](https://github.com/edanalytics/edu_wh/tree/feature/discipline_updates) or else we'll run into some merge conflicts later on (unless there is a more urgent need to get merged)

susanxiong:
> @susanxiong the changes here look good, we'll probably want to combine this with the [other discipline branch](https://github.com/edanalytics/edu_wh/tree/feature/discipline_updates) or else we'll run into some merge conflicts later on (unless there is a more urgent need to get merged)

@jalvord1 what's the timeline for getting that other branch merged?

models/core_warehouse/fct_student_discipline_actions.sql
```diff
@@ -95,6 +95,11 @@ join_descriptor_interpretation as (
         xwalk_discipline_actions.is_iss,
         xwalk_discipline_actions.is_exp,
         xwalk_discipline_actions.is_minor,
+
+        -- bring in any additional custom columns from xwalk, does nothing if there are no extra columns
+        {{ accordion_columns('xwalk_discipline_actions',
+            exclude_columns=['discipline_action','severity_order','is_oss','is_iss','is_exp','is_minor']) }}
```
rlittle08, line 101: if any of these columns e.g. `is_minor` are missing from the xwalk, I think this code this will break. I doubt that would happen in any of our implementations, but since it's config they technically could be removed. maybe we should update accordion_columns to ignore when an `exclude_column` doesn't exist?

rlittle08: (I don't think that should stop us from merging this, though)

susanxiong: good call. if those columns are missing from the xwalk, this model would break regardless (even without accordion_columns) because all of the "default" xwalk columns are included in the model (lines 94-97). but we can add a to-do item in the edu project development board to investigate how accordion_columns will behave if an excluded column doesn't exist.

susanxiong: i tested it out and actually the macro that accordion_columns uses will just ignore a column if it doesn't exist. https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/get_filtered_columns_in_relation.sql

PR #73, opened 2023-09-08
add configurable stu-program-agg columns

rlittle08:
## Description and Motivation
By default in EDU, programs have `is_active` and `is_annual` as indicators that are aggregated by `k_student`. Those are written in build models like [here](https://github.com/edanalytics/edu_wh/blob/239d4b3acdf56a707dea04b4c4bd860c28ad7fd8/models/build/edfi_3/students/bld_ef3__student_program__special_education.sql#L22). This feature branch adds the ability to add arbitrarily many more of those types of student aggregate program indicators, with custom rules.

For example, if you want to add a new indicator for whether a student was in a 504 Plan Program during the school year, you can configure these variable in your dbt_project.yml, assuming your implementation has `program_name` values equal to '504 Plan':

```
  # custom indicators in special ed model at the same agg level as is_special_ed_active & annual
  'edu:special_ed:custom_program_agg_indicators':
    is_504_plan_annual:
      agg_sql: >
        -- if the student had a 504 program enrollment any time during the year  
 
        max(
          program_name = '504 Plan'
        )
```

And the compiled code in the build model will look like this:
```
-- custom special ed program agg indicators
-- if the student had a 504 program enrollment any time during the year
max(
  program_name = '504 Plan'
)
 as is_504_plan_annual,
```

And any student with a program enrollment with name '504 Plan' will now have `is_504_plan_annual` = TRUE in `dim_student`. Plus, those values will be automatically added to `dim_subgroup` and `fct_student_subgroup`.

- `bld_ef3__student_program__special_education`, this code optionally adds more agg columns by grabbing sql code from the project config:
https://github.com/edanalytics/edu_wh/blob/543a8b6d75e1984d5be85f8f5ffbfb8e5561f258/models/build/edfi_3/students/bld_ef3__student_program__special_education.sql#L36-L41
  - (Same code is repeated for `bld_ef3__student_program__homeless`, `bld_ef3__student_program__language_instruction`, and `bld_ef3__student_program__title_i` )
- `dim_student`, this code adds new custom indicators as columns: https://github.com/edanalytics/edu_wh/blob/543a8b6d75e1984d5be85f8f5ffbfb8e5561f258/models/core_warehouse/dim_student.sql#L93-L96
  - (Same code is repeated for `bld_ef3__student_program__homeless`, `bld_ef3__student_program__language_instruction`, and `bld_ef3__student_program__title_i` )

## New models added:
**None**

## Tests and QC Done:
Tested on Boston with configuration of 504 plan annual & active indicators.

Also tested on SCDE with configuration of homeless unaccompanied youth indicators.

Tested in both Boston and SCDE with an empty & missing config

### If adding configuration, did you test with empty, broken, and nonexistent configuration?
Yes, I tested with no configuration added, and the new code compiles to empty. 

## PR Merge Priority:
(Promised this feature by end of Sept for BPS)
- [ ] Low
- [ ] Medium
- [x] High

## Questions:
This code assumes your configuration will include SQL to aggregate over `k_student`. What if you want to add a `k_student_xyear` aggregation? Should that be included in this branch?

### DIFF
--- a/models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
+++ b/models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
@@ -33,7 +33,7 @@ maxed as (
         ) as is_homeless_annual, -- the student had a homeless program enrollment any time during the year
 
         -- custom homeless program agg indicators
-        {% if custom_program_agg_indicators is not none and custom_program_agg_indicators | length -%}
+        {% if custom_program_agg_indicators -%}
           {%- for indicator in custom_program_agg_indicators -%}
             {{ custom_program_agg_indicators[indicator]['agg_sql'] }} as {{ indicator }},
           {%- endfor -%}
--- a/models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
+++ b/models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
@@ -7,6 +7,8 @@
 {# customizable: the column that defines the end date for the homeless program #}
 {% set exit_date_column = var('edu:homeless:exit_date_column') %}
 
+{# customizable: extra indicators to create in the aggregate query #}
+{% set custom_program_agg_indicators = var('edu:homeless:custom_program_agg_indicators', None) %}
 
 with stage as (
     select * from {{ ref('stg_ef3__student_homeless_program_associations') }}
@@ -29,6 +31,13 @@ maxed as (
         max(
           {{ value_not_in_list(field='program_name', excluded_items=exclude_programs) }}
         ) as is_homeless_annual, -- the student had a homeless program enrollment any time during the year
+
+        -- custom homeless program agg indicators
+        {% if custom_program_agg_indicators is not none and custom_program_agg_indicators | length -%}
+          {%- for indicator in custom_program_agg_indicators -%}
+            {{ custom_program_agg_indicators[indicator]['agg_sql'] }} as {{ indicator }},
+          {%- endfor -%}
+        {%- endif %}
 
         max(is_awaiting_foster_care) as is_awaiting_foster_care,
         max(is_homeless_unaccompanied_youth) as is_homeless_unaccompanied_youth,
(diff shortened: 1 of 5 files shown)


jalvord1 (commented):
we talked through this idea but noting here that I think it's a common enough use-case to account for any additional complexity, especially considering the added logic to not require the new var

models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
```diff
@@ -30,6 +32,13 @@ maxed as (
           {{ value_not_in_list(field='program_name', excluded_items=exclude_programs) }}
         ) as is_homeless_annual, -- the student had a homeless program enrollment any time during the year
 
+        -- custom homeless program agg indicators
+        {% if custom_program_agg_indicators is not none and custom_program_agg_indicators | length -%}
```
jayckaiser, line 36: FYI, I believe we can simplify this condition to the following:
`{% if custom_program_agg_indicators %}`

models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
```diff
@@ -30,6 +32,13 @@ maxed as (
...
+          {%- for indicator in custom_program_agg_indicators -%}
+            {{ custom_program_agg_indicators[indicator]['agg_sql'] }} as {{ indicator }},
```
jayckaiser, line 38: What is the reason for nesting the SQL query under the `agg_sql` key? Are there other keys that we will want to add in the future for flexibility?

jayckaiser:
I've left two comments regarding general structure of the changes. Overall, this addition looks good and adds much-needed flexibility when it comes to custom programs.

models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
rlittle08: nice, just removed the redundant code everywhere

models/build/edfi_3/students/bld_ef3__student_program__homeless.sql
rlittle08: it is possible in future we want more attributes, e.g. `include_in_dim_student`

ejoranlienea (approved):

PR #78, opened 2023-10-20
Feature/add warehouse models for cohorts

sleblanc23:
## Description & motivation
Adds warehouse models for cohorts, per the [EDU Project Development](https://educationanalytics.monday.com/boards/3556827529/pulses/5009936520) board.

Requires a field added to `stg_ef3__cohorts` in this [PR.](https://github.com/edanalytics/edu_edfi_source/pull/46)

## PR Merge Priority:
- [x] Low

## New files created:
- `dim_cohort.sql` 
- `fct_student_cohort_association.sql`

## Tests and QC done:
- Ran the models successfully in GSN and South Carolina

## Future ToDos:
No implementations have data for staffCohortAssociations yet. I added a [Monday task](https://educationanalytics.monday.com/boards/3556827529/pulses/5305893276) to build `fct_staff_cohort_association` once we have a use case.

### DIFF
--- a/models/core_warehouse/dim_cohort.sql
+++ b/models/core_warehouse/dim_cohort.sql
@@ -1,0 +1,28 @@
+{{
+  config(
+    post_hook=[
+        "alter table {{ this }} add primary key (k_cohort)",
+    ]
+  )
+}}
+
+with stg_cohorts as (
+    select * from {{ ref('stg_ef3__cohorts') }}
+),
+formatted as (
+    select 
+        stg_cohorts.k_cohort,
+        stg_cohorts.k_lea,
+        stg_cohorts.k_school,
+        stg_cohorts.tenant_code,
+        stg_cohorts.school_year,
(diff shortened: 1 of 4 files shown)


ejoranlienea:
Looks good. 
Two thoughts:
1. Could consider adding an active indicator, like [student_school_association has](https://github.com/edanalytics/edu_wh/blob/1f7567625cd9439073e5d91ffff0c6bfea575886/models/core_warehouse/fct_student_school_association.sql#L54)
2. Given how rarely (if ever) we've seen this data in the wild, I wonder if we should tag this under its own domain here and in source so that we can disable it when not in use.

sleblanc23:
@ejoranlienea 

I added an active indicator and a `cohort` tag so we can disable these when they're not needed. Also opened a [PR in source](https://github.com/edanalytics/edu_edfi_source/pull/48) for those models. Am I missing anything for adding that tag?

ejoranlienea (approved):

PR #79, opened 2023-10-30
Feature/optional filter early exits

rlittle08:
## Description & motivation
Currently, EDU excludes students who exited before the first day of school from `fct_student_school_association`. This PR makes that rule optional. The existing code is there to avoid carrying around student records that aren't relevant for analysis, but in certain cases, these students _are_ needed for analysis, e.g. dropout analysis. 

This PR is non-breaking, non-changing by default, but if implementations add variable config:
```
'edu:enroll:exclude_exit_before_first_day': False
```
, then early-exit students will be included in the warehouse on future dbt runs.

## PR Merge Priority:

- [x] Low
- [ ] Medium
- [ ] High

## Changes to existing files:

- `fct_student_school_association`: add conditional block around existing exclusion code

## New files created:

None

## Tests and QC done:

Tested on Boston with empty, False, and True config

## Future ToDos & Questions:

- Should early exit students be tracked in a separate table?

### DIFF
--- a/models/core_warehouse/fct_student_school_association.sql
+++ b/models/core_warehouse/fct_student_school_association.sql
@@ -92,8 +92,7 @@ formatted as (
         and stg_stu_school.school_year = bld_school_calendar_windows.school_year
         and equal_null(dim_school_calendar.k_school_calendar, bld_school_calendar_windows.k_school_calendar)
     where true
-   {% set excl_early_exit =  var('edu:enroll:exclude_exit_before_first_day', True)  %}
-   {% if excl_early_exit -%}
+   {% if var('edu:enroll:exclude_exit_before_first_day', True) -%}
       -- exclude students who exited before the first school day
       and (exit_withdraw_date >= bld_school_calendar_windows.first_school_day
           or exit_withdraw_date is null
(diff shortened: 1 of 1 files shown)


models/core_warehouse/fct_student_school_association.sql
```diff
@@ -92,11 +92,14 @@ formatted as (
         and stg_stu_school.school_year = bld_school_calendar_windows.school_year
         and equal_null(dim_school_calendar.k_school_calendar, bld_school_calendar_windows.k_school_calendar)
     where true
-    -- exclude students who exited before the first school day
-    and (exit_withdraw_date >= bld_school_calendar_windows.first_school_day
-        or exit_withdraw_date is null
-        or bld_school_calendar_windows.first_school_day is null
-        )
+   {% set excl_early_exit =  var('edu:enroll:exclude_exit_before_first_day', True)  %}
```
ejoranlienea, line 95: Is this `set` necessary, or can it just be `if var(...)`?

ejoranlienea (changes requested):

models/core_warehouse/fct_student_school_association.sql
rlittle08: done & tested

ejoranlienea (approved):

PR #84, opened 2023-11-10
Allow custom source for student daily attendance

susanxiong:
## Description & motivation
These code changes allow us to swap out the source data behind `fct_student_daily_attendance` while retaining that fact model and any current or future downstream models built off of it (e.g., `msr_student_cumulative_attendance`).

Non-breaking change. If no custom model is specified in the dbt_project.yml file, then it will default to the original source: `fct_student_school_attendance_event`.

Allow for a custom source for all three of these models that reference fct_student_school_attendance_event
- models/build/edfi_3/bld_ef3__attendance_sessions.sql
- models/core_warehouse/fct_student_daily_attendance.sql
- models/qc/attendance_freshness.sql

Add an optional variable to include a custom docs page in fct_student_daily_attendance.yml

## New models created:
none

## Tests and QC done:
Tested in Jeffco with and without a custom source specified in the dbt_project.yml file. Runs successfully.

## PR Merge Priority:
High

### DIFF
--- a/models/build/edfi_3/bld_ef3__attendance_sessions.sql
+++ b/models/build/edfi_3/bld_ef3__attendance_sessions.sql
@@ -10,7 +10,7 @@ with dim_session as (
     select * from {{ ref('dim_session') }}
 ),
 fct_student_sch_attend as (
-    select * from {{ ref('fct_student_school_attendance_event') }}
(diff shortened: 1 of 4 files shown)


susanxiong:
more info to come after i connect with Sam next week

sleblanc23 (approved):

ejoranlienea (approved):

PR #88, opened 2023-11-20
Bugfix/attendance calendars from bell schedules

sleblanc23:
## Description & motivation
Addresses an edge case in the `fct_student_daily_attendance` model. Students who only attend school part-time (only on Tuesday/Thursday, for example) are currently shown to be in attendance on days they don't go to school. This PR fixes that by constructing the calendar of possible attendance days for each student based on their class schedules. 

## Changes to existing files:
- `fct_student_daily_attendance` : Replace the calendar-constructing steps with a join to the newly added build model.

## New files created:
- `bld_ef3__student_school_days` : Constructs the student-school level instructional calendars, including logic that was previously in `fct_student_daily_attendance` for removing non-instructional days and any that fall outside the student's enrollment. Put this in a separate model to keep it readable given of the large number of joins required.

## Tests and QC done:
Confirmed that it fixed the identified edge cases without creating any new inaccuracies for three schools in Jeffco. Additional QC to be done before marking ready for review.

## Future ToDos & Questions:
It's possible that this creates more problems than it solves when bell schedule data is bad or incomplete.

### DIFF
--- a/models/build/edfi_3/bld_ef3__student_school_days.sql
+++ b/models/build/edfi_3/bld_ef3__student_school_days.sql
@@ -1,0 +1,85 @@
+with student_school_enr as (
+    select * from {{ ref('fct_student_school_association') }}
+),
+student_section_enr as (
+    select * from {{ ref('fct_student_section_association') }}
+),
+course_section as (
+    select * from {{ ref('dim_course_section') }}
+),
+section_class_periods as (
+    select * from {{ ref('stg_ef3__sections__class_periods') }}
+),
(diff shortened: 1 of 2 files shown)


ejoranlienea (commented):
Would anything change about this object if the following joins were removed?
- course_section
- class_period
- bell

It looks like in each of these cases, the subsequent joins use the same keys that are already present and nothing else from these objects is used directly, so we may be able to trim this down by a few joins without affecting anything.

PR #89, opened 2023-11-27
initial draft

tomreitz:
This adds a model `fct_network_association` which allows for association of a school or LEA to multiple networks. Mainly we're adding this to enable easier network-level permissioning.


rlittle08:
we got an interesting request from Boston about potentially storing network associations by year (so if school leaves network, that information is stored annually). I'm not sure if that would be part of this table, but maybe?

PR #94, opened 2023-12-07
choose latest entry date rather than longest

ejoranlienea:


models/build/edfi_3/students/bld_ef3__stu_grade_level.sql
```diff
@@ -11,10 +11,12 @@ find_grade_level as (
     qualify 1 = row_number() over(
         partition by k_student, school_year
         order by 
-            -- days enrolled, descending
+            -- latest entry date
+            entry_date desc
```
rlittle08, line 15: missing trailing comma

models/build/edfi_3/students/bld_ef3__stu_grade_level.sql
```diff
@@ -11,10 +11,12 @@ find_grade_level as (
...
+            -- tie break on longer
             (coalesce(exit_withdraw_date, current_date()) - entry_date) desc,
```
rlittle08, line 17: since we have entry_date above, could change this to just ```exit_withdraw_date desc nulls first``` to get same result, and might be quicker? but probably comparable

rlittle08 (changes requested):

PR #96, opened 2023-12-13
docs/fct_student_assessment

sleblanc23:
Added a sample query for unpacking a particular score result from `fct_student_assessment.v_other_results`. Denver is interested in several of these score results for their reporting and I thought it was worth adding the sample here rather than in a Denver custom doc.

### DIFF
--- a/models/core_warehouse/fct_student_assessment.yml
+++ b/models/core_warehouse/fct_student_assessment.yml
@@ -10,6 +10,19 @@ models:
       ##### Primary Key:
         `k_student_assessment` -- There is one record per student-assessment event
 
+      ##### Example of querying "other results":
+      Any score that are not included in the wide score results columns are listed in the `v_other_results` column. They can be included
+      as wide columns in a query like this:
+      ```
+          SELECT 
+            fsa.*, 
+            fsa.v_other_results:"{NAME OF SCORE}"::int as name_of_score
+          FROM analytics.prod_wh.fct_student_assessment fsa
+          JOIN analytics.prod_wh.dim_assessment da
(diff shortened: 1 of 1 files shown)


models/core_warehouse/fct_student_assessment.yml
```diff
@@ -10,6 +10,19 @@ models:
       ##### Primary Key:
         `k_student_assessment` -- There is one record per student-assessment event
 
+      ##### Example of querying "other results":
```
rlittle08, line 13: very picky but can we use the ```##### Example Use Cases:``` header instead, to be consistent with other docs pages?

rlittle08 (changes requested):

models/core_warehouse/fct_student_assessment.yml
sleblanc23: fixed!

rlittle08 (approved):

models/core_warehouse/fct_student_assessment.yml
```diff
@@ -10,6 +10,19 @@ models:
       ##### Primary Key:
         `k_student_assessment` -- There is one record per student-assessment event
 
+      ##### Example Use Cases:
+      Any score that are not included in the wide score results columns are listed in the `v_other_results` column. They can be included
```
jalvord1, line 14: this is tiny, can we change to 'Any scores' or 'Any score that is'?

sleblanc23: oops. fixed!

jalvord1 (approved):

ejoranlienea (approved):

PR #95, opened 2023-12-11
Student academic records - k_student updates

sleblanc23:
## Description & motivation
`fct_student_academic_record` and its dependent models currently include `k_student` only, which excludes any historical records from years prior to the earliest `dim_student` records. This PR brings in `k_student_xyear` and updates the logic to allow `k_student` to be null.

## Changes to existing files:
- `fct_course_transcripts`, `fct_student_academic_record`, `fct_student_gpa` : Added `k_student_xyear` and `k_student` via a left join to `dim_student`; updated constraints
- yml files for these models: Updated to include a sample query for joining to the most recent `dim_student` record

## Tests and QC done (together with [edu_edfi_source):](https://github.com/edanalytics/edu_edfi_source/pull/55)
- Confirmed that row counts were unchanged in Jeffco and correct k_student was present 

## PR Merge Priority:
Low - note this depends on this [edu_edfi_source](https://github.com/edanalytics/edu_edfi_source/pull/55) PR

PR #100, opened 2024-01-05
fix/historic assessments

rlittle08:
## Description & motivation
To align with similar branch for stu academic records https://github.com/edanalytics/edu_wh/pull/95 -- Make `k_student` null in `fct_student_assessment` for records where there's no contemporaneous (stole that word from Erik) dim_student record. Also, remove records if there is no record for the given student in any year (using a lookup via k_student_xyear).

This enables downstream apps to query a full history of student assessment records. But in order to attach them to correct demographics, they'll need to be careful with their query. I made note of this in dbt docs, but should we add an example query, or better yet, a view for this?

## PR Merge Priority:
- [ ] Low
- [x] Medium (for New Rally)
- [ ] High

## Tests and QC done:
Tested on both TX Exchange (dev_raw has historic assessments for dallas) and SC Districts (prod raw has historic SCREADY). `k_student` values are made null for historic years as expected, and rows are dropped for students missing from known demographics.

A quick performance check in SC showed small change for fct_student_assessment (18.03s vs. 18.67s on my laptop).

## Future ToDos & Questions:
- Should we include a convenience view here for those who want a prebuilt demographics join?

### DIFF
--- a/models/core_warehouse/fct_student_assessment.sql
+++ b/models/core_warehouse/fct_student_assessment.sql
@@ -29,9 +29,9 @@ student_assessments_wide as (
     select
         student_assessments.k_student_assessment,
         student_assessments.k_assessment,
-        student_assessments.k_student_xyear,
         -- use dim_student.k_student. NOTE, will be null when no corresponding demographics found (e.g. historic year of assessment data)
         dim_student.k_student,
+        student_assessments.k_student_xyear,
         student_assessments.tenant_code,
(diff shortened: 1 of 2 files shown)


jalvord1 (approved):
@ejoranlienea think we can get this one through before Jan 31? Need for new rally. I know you're putting together 0.3.0 just can't remember what you said about timing on that one.

ejoranlienea (approved):

PR #110, opened 2024-02-06
measure attendance freshness by school

ejoranlienea:
Our original thought was that attendance would get blocked up at the SIS/District-level, but in practice there is a lot of variance by school, with some schools falling quite a bit behind the rest of the district.

Breaking this down to the school level should make this more useful for diagnostics. This is still compatible with the downstream test, which is agnostic about the columns in this table.

### DIFF
--- a/models/qc/attendance_freshness.sql
+++ b/models/qc/attendance_freshness.sql
@@ -12,3 +12,4 @@ from att_events
 join dim_calendar
     on att_events.k_calendar_date = dim_calendar.k_calendar_date
 group by all
+having max_date <= current_date()
--- a/models/qc/attendance_freshness.sql
+++ b/models/qc/attendance_freshness.sql
@@ -11,5 +11,5 @@ select
 from att_events
 join dim_calendar
     on att_events.k_calendar_date = dim_calendar.k_calendar_date
+where calendar_date <= current_date()
 group by all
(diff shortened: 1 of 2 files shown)


rlittle08:
tested this in boston, looks good, but ryan looking into an edge case re: future attendance events

models/qc/attendance_freshness.sql
```diff
@@ -6,8 +6,10 @@ dim_calendar as (
...
     att_events.tenant_code,
+    att_events.k_school,
     max(calendar_date) as max_date
 from att_events
 join dim_calendar
     on att_events.k_calendar_date = dim_calendar.k_calendar_date
-group by 1
\ No newline at end of file
+group by all
+having max_date <= current_date()
```
rlittle08, line 15: @ryanaguilar I think by doing a `having`, if there are future dates, the school will be left out of the test. What I was thinking was to filter using a `where`, so the school is included, but using the max date `where not in future`

rlittle08: I resolved this^

rlittle08:
@ejoranlienea In boston this produces some cases of schools that closed in prior years, or summer schools. So I'm going to add a filter to latest school year for this test

rlittle08 (approved):

PR #113, opened 2024-02-09
Feature/custom override grade level

jayckaiser:
# Feature: Custom-Override Grade Level

## Description & motivation
In South Carolina, students are assigned an **enrolled grade** and **true grade**. These usually align, but there are cases where a student's courses are for a grade that differs from their enrollment (e.g., True Grade Z1 = repeating 12th grade for the first time while taking 12th grade classes).

South Carolina has requested overriding enrollment grade in cases like these, but we have not built a mechanism in the warehouse to complete this logic. This PR introduces a DBT variable that specifies a `source` and a `where` statement to override grade level when applicable.

(Alternatively, we can include a True Grade column in `dim_student`, but that may introduce confusion for end-users.)

## PR Merge Priority:

- [x] Low
- [ ] Medium
- [ ] High

## New files created:
- `bld_ef3__stu_grade_level`: Add CTE that checks for the new variable and coalesces its value over grade level when defined.
- `dbt_project.yml`: Add new variable, defaulting to Null.

## Tests and QC done:
This has been run successfully in dev on South Carolina using their TrueGrade SEOA student indicator.


rlittle08 (approved):
added a test for uniqueness, but LGTM for v0.3.2

PR #115, opened 2024-03-01
fct_stu_contact_information

keenzarate213:


models/core_warehouse/fct_student_contact_information.sql
```diff
@@ -0,0 +1,39 @@
...
+    congressional_district,
+    county_fips_code,
+    latitude,
+    longitude,
+    email_type, 
+    email_address,
+    phone_number_type, 
+    phone_number
+from dim_student 
+left join stg_stu_address on dim_student.k_student = stg_stu_address.k_student
```
rlittle08, line 37: these joins blow up the grain, combining every student x every address they have x every email they have etc... we should instead make address, email, etc. wide before joining

models/core_warehouse/fct_student_contact_information.sql
```diff
@@ -0,0 +1,39 @@
...
+    latitude,
+    longitude,
+    email_type, 
+    email_address,
+    phone_number_type, 
+    phone_number
+from dim_student 
+left join stg_stu_address on dim_student.k_student = stg_stu_address.k_student 
+left join stg_stu_emails  on dim_student.k_student = stg_stu_emails.k_student 
+left join stg_stu_phones  on dim_student.k_student = stg_stu_phones.k_student
```
rlittle08, line 39: potentially add language as a contact field

models/core_warehouse/dim_student.sql
```diff
@@ -50,6 +50,13 @@ stu_grade as (
     select * from {{ ref('bld_ef3__stu_grade_level') }}
 ),
 
+choose_language as (
```
rlittle08, line 53: make this a build model to hide the complexity outside of dim_student

models/core_warehouse/dim_student.sql
```diff
@@ -50,6 +50,13 @@ stu_grade as (
...
+    {{ row_pluck(ref('stg_ef3__stu_ed_org__languages'),
+                key='k_student',
+                column='language_use',
+                preferred='Home language') }}
```
rlittle08, line 57: once in build model, the preferred argument should be configurable, but can default to home language

models/core_warehouse/dim_student.sql
```diff
@@ -188,8 +195,8 @@ formatted as (
         stg_student.api_year = max(stg_student.api_year) over(partition by stg_student.k_student_xyear) as is_latest_record,
        
         stu_immutable_demos.race_array,
-        stu_immutable_demos.safe_display_name
-
+        stu_immutable_demos.safe_display_name,
+        choose_language.code_value as home_language
```
rlittle08, line 199: don't hard code home_language, make configurable

rlittle08 (changes requested):

models/build/edfi_3/students/bld_ef3__student_wide_address.sql
```diff
@@ -0,0 +1,27 @@
+with stg_student_address as (
+    select * from {{ ref('stg_ef3__stu_ed_org__addresses') }}
+),
+address_types as (
+    select * from {{ ref('xwalk_student_address_types') }}
```
rlittle08, line 5: I think instead of adding these xwalks as another thing to manage, we could just run `get_column_values()` on the values from the stg table. here's a working draft that I could commit to the branch if you'd like

```
with stg_student_address as (
    select * from {{ ref('stg_ef3__stu_ed_org__addresses') }}
),
address_wide as (
  select 
    k_student,
    tenant_code
    {%- if not is_empty_model('stg_ef3__stu_ed_org__addresses') -%},
    {{ dbt_utils.pivot(
      'address_type',
      dbt_utils.get_column_values(ref('stg_ef3__stu_ed_org__addresses'), 'address_type'),
      agg='max',
      then_value='street_address',
      else_value='null',
      suffix='_address',
      quote_identifiers = False
    ) }}
    {%- endif %}
  from stg_student_address
  group by k_student, tenant_code
)
select * from address_wide
```

rlittle08 (changes requested):

rlittle08:
one major TODO: how to deal with the potential case that e.g. address types & language types might overlap? Currently, this would cause duplicate colnames in `fct_student_contact_information`, which causes SQL error. I can't yet think of a quick way to handle this case. 

A similar but less threatening case is e.g. language types that aren't specific enough. One of the Ed-Fi defaults for Language Use is  'Other', and this is pop'd in Boston, so `fct_student_contact_information` ends up with a column just called `other`. Not descriptive. Should we have some way of creating configurable aliases in the `bld_ef3__student_wide..` tables, which might resolve both these issues?

rlittle08:
Marking this back to draft while we figure out best path forward

PR #118, opened 2024-03-20
Feature/assessment sy

jalvord1:
## Description & motivation
School year is not a required field in the student assessment entity of Ed-Fi, but it is commonly used/necessary for certain reporting, including Podium. This feature adds a customizable school year calculation using administration dates.

## Changes to existing files:
- `fct_student_assessment` : Adds a coalesce around `school_year` with a calculation using `administration_date` from either a macro call or a new xwalk.
- `fct_student_objective_assessment` : Same as above.
- `dbt_project`: Adds default values for school year start day and month.

## New files created:
- `derive_school_year` : This macro is broadly written to be used for any case where a school year needs to be calculated from a date. The logic uses the variables `school_year:start_month` and `school_year:start_day` to test whether a date falls within the school year of that date, or year + 1.

## Tests and QC done:
I tested this feature with a few different scenarios:
- Using the default variable values set in `edu_wh` without the xwalk variable and without the xwalk existing in the implementation repo. The code ran successfully and filled in school years correctly when they were null. 
- Using new variable values set in the implementation repo. The code ran successfully and filled in school years correctly (using the new variable date cutoffs) when they were null. 
- Using the xwalk (& setting the xwalk variable to true). I left the `assessment_identifier` and `namespace` columns null to test the optional join on those columns. The code ran successfully and filled in school years correctly when they were null. 
- Using the xwalk (& setting the xwalk variable to true). I filled in the `assessment_identifier` and `namespace` columns with an assessment that _never_ has null school years. The code ran successfully and none of the school years were overwritten. 
- Using the xwalk (& setting the xwalk variable to true). I filled in the `assessment_identifier` and `namespace` columns with an assessment that _only_ has null school years. The code ran successfully and filled in school years correctly when they were null. 

## Future ToDos & Questions:
See comment below with drawbacks.

## PR Merge Priority:
- Medium

### DIFF
--- a/macros/derive_school_year.sql
+++ b/macros/derive_school_year.sql
@@ -1,19 +1,10 @@
 {# 
 This macro takes derives school years from a date column.
+
+The date column will be compared to the default school year start day and month.
 #}
 
 {% macro derive_school_year(date_column) %}
-
-    {# {% set date_lower_bound = 
-            "concat_ws('/', '{{var('edu:school_year:start_month')}}', 
-                           '{{var('edu:school_year:start_day')}}', 
-                           year(date_column::date))::date" %}
-
-    {% set date_upper_bound = 
-            "concat_ws('/', '{{var('edu:school_year:start_month')}}', 
-                           '{{var('edu:school_year:start_day')}}', 
-                           (year(date_column::date)::int + 1))::date" %} #}
-
     case
         when {{date_column}}::date >= 
             concat_ws('/', '{{var("edu:school_year:start_month")}}', 
@@ -26,5 +17,4 @@ This macro takes derives school years from a date column.
             then (year({{date_column}}::date)::int + 1)
         else year({{date_column}}::date)
     end
-
 {% endmacro %}
--- a/macros/derive_school_year.sql
+++ b/macros/derive_school_year.sql
@@ -5,6 +5,7 @@ The date column will be compared to the default school year start day and month.
 #}
 
 {% macro derive_school_year(date_column) %}
+    {% if var("edu:school_year:start_month", None) and var("edu:school_year:start_day", None) %}
     case
         when {{date_column}}::date >= 
             concat_ws('/', '{{var("edu:school_year:start_month")}}', 
@@ -17,4 +18,7 @@ The date column will be compared to the default school year start day and month.
             then (year({{date_column}}::date)::int + 1)
         else year({{date_column}}::date)
     end
+    {% else %}
+    null
+    {% endif %}
 {% endmacro %}
--- a/macros/derive_school_year.sql
+++ b/macros/derive_school_year.sql
@@ -1,0 +1,30 @@
+{# 
+This macro takes derives school years from a date column.
+#}
+
+{% macro derive_school_year(date_column) %}
+
+    {# {% set date_lower_bound = 
+            "concat_ws('/', '{{var('edu:school_year:start_month')}}', 
+                           '{{var('edu:school_year:start_day')}}', 
+                           year(date_column::date))::date" %}
+
+    {% set date_upper_bound = 
+            "concat_ws('/', '{{var('edu:school_year:start_month')}}', 
+                           '{{var('edu:school_year:start_day')}}', 
+                           (year(date_column::date)::int + 1))::date" %} #}
+
+    case
+        when {{date_column}}::date >= 
+            concat_ws('/', '{{var("edu:school_year:start_month")}}', 
+                           '{{var("edu:school_year:start_day")}}', 
+                           year({{date_column}}::date))::date
+            and {{date_column}}::date < 
+            concat_ws('/', '{{var("edu:school_year:start_month")}}', 
+                           '{{var("edu:school_year:start_day")}}', 
+                           (year({{date_column}}::date)::int + 1))::date
+            then (year({{date_column}}::date)::int + 1)
+        else year({{date_column}}::date)
+    end
+
+{% endmacro %}
--- a/models/core_warehouse/fct_student_assessment.sql
+++ b/models/core_warehouse/fct_student_assessment.sql
@@ -76,7 +76,7 @@ student_assessments_wide as (
     {% if var('edu:school_year:assessment_dates_xwalk_enabled', False) %}
     left join {{ ref('xwalk_assessment_school_year_dates') }} dates_xwalk
         -- note: between means A >= X AND A <= Y, so date upper/lower bounds should not overlap across years
-        on student_assessments.administration_date between start_date::date and end_date::date
+        on student_assessments.administration_date between dates_xwalk.start_date::date and dates_xwalk.end_date::date 
         -- we want to allow for the school year cutoffs to differ by assessment 
         -- but also allow those fields to remain null if xwalk is desired but not to differ across assessments
         and ifnull(dates_xwalk.assessment_identifier, '1') = iff(dates_xwalk.assessment_identifier is null, '1', student_assessments.assessment_identifier)
--- a/models/core_warehouse/fct_student_assessment.sql
+++ b/models/core_warehouse/fct_student_assessment.sql
@@ -35,8 +35,11 @@ student_assessments_wide as (
         student_assessments.tenant_code,
         student_assessments.student_assessment_identifier,
         student_assessments.serial_number,
-        {% if var('edu:school_year:dates_xwalk_enabled', False) %}
-        coalesce(student_assessments.school_year, dates_xwalk.school_year, {{derive_school_year('student_assessments.administration_date')}}) as school_year,
+        {% if var('edu:school_year:assessment_dates_xwalk_enabled', False) %}
+        iff(dates_xwalk.override_existing,
+            coalesce(dates_xwalk.school_year, student_assessments.school_year, {{derive_school_year('student_assessments.administration_date')}}),
+            coalesce(student_assessments.school_year, dates_xwalk.school_year, {{derive_school_year('student_assessments.administration_date')}}))
+        as school_year,
         {% else %}
         coalesce(student_assessments.school_year, {{derive_school_year('student_assessments.administration_date')}}) as school_year,
         {% endif %}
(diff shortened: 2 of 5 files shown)


jalvord1:
Some drawbacks I've identified (some more important than others):
1. The optional join of both `assessment_identifier` and `namespace` could allow for someone to only fill in one of those and the code would still successfully run. For example, If two _different_ assessments had the same `assessment_identifier` and only that column was filled in, those dates would apply to both assessments. This feels low risk, as both assessments in that case would need to have null school years but different desired date cutoffs for this to even apply or be a problem.
2. The `coalesce` would mean that there is no way to avoid filling in null schools years if they exist. I think this is desired, but pointing out regardless.
3. I noted this in the code but want to call out here as well: the join of dates using `between` requires that date upper/lower bounds *not* overlap. If across years the upper/lower bounds _did_ overlap, then undesired behavior could occur.
4. This is also noted as a todo in the code: Because some of the logic in this feature can be used beyond this use case, I tried to keep certain aspects assessment-agnostic, including the `dates_xwalk_enabled` variable. However, we will need separate date xwalks for different domains because we are allowing for domain-specific rules (like different date cutoffs by assessment, for example). This means we should likely have separate variables, like `assessment_dates_xwalk_enabled`, but that leads to variable bloat.

models/core_warehouse/fct_student_assessment.sql
```diff
@@ -35,7 +35,11 @@ student_assessments_wide as (
         student_assessments.tenant_code,
         student_assessments.student_assessment_identifier,
         student_assessments.serial_number,
-        student_assessments.school_year,
+        {% if var('edu:school_year:dates_xwalk_enabled', False) %}
+        coalesce(student_assessments.school_year, dates_xwalk.school_year, {{derive_school_year('administration_date')}}) as school_year,
+        {% else %}
+        coalesce(student_assessments.school_year, {{derive_school_year('administration_date')}}) as school_year,
```
rlittle08, line 41: Is there any world where we should allow an implementation to keep school_year NULL? instead of forcing to using 8/1 or configure different dates? (this is exactly your point 2., I think)

models/core_warehouse/fct_student_assessment.sql
```diff
@@ -35,7 +35,11 @@ student_assessments_wide as (
         student_assessments.tenant_code,
         student_assessments.student_assessment_identifier,
         student_assessments.serial_number,
-        student_assessments.school_year,
+        {% if var('edu:school_year:dates_xwalk_enabled', False) %}
```
rlittle08, line 38: potentially help your point 4., could we instead use `if not is_empty_model(xwalk)`, so if xwalk is filled in, it's used without needing to add this var? I suppose that denies ability to fill in xwalk & then turn it off

models/core_warehouse/fct_student_assessment.sql
```diff
@@ -66,6 +70,16 @@ student_assessments_wide as (
     -- left join to allow 'historic' records (assess records with no corresponding stu demographics)
     left join dim_student
         on student_assessments.k_student = dim_student.k_student
+    -- todo: if we are going to have multiple xwalks, should there be multiple vars? specific to assessment for ex.
+    {% if var('edu:school_year:dates_xwalk_enabled', False) %}
+    left join {{ ref('assessment_school_year_dates') }} dates_xwalk
+        -- note: between means A >= X AND A <= Y, so date upper/lower bounds should not overlap across years
```
rlittle08, line 76: to address your point 3., maybe we should make a test for overlapping dates? if you agree I could draft that

macros/derive_school_year.sql
```diff
@@ -0,0 +1,20 @@
+{# 
+This macro takes derives school years from a date column.
+
+The date column will be compared to the default school year start day and month.
+#}
+
+{% macro derive_school_year(date_column) %}
+    case
+        when {{date_column}}::date >= 
+            concat_ws('/', '{{var("edu:school_year:start_month")}}',
```
rlittle08, line 10: can we add some error handling for invalid month/days? e.g. if I put start_month = "50" the model errors. maybe that's desirable though, to error loudly?

models/core_warehouse/fct_student_assessment.sql
```diff
@@ -35,7 +35,11 @@ student_assessments_wide as (
         student_assessments.tenant_code,
         student_assessments.student_assessment_identifier,
         student_assessments.serial_number,
-        student_assessments.school_year,
+        {% if var('edu:school_year:dates_xwalk_enabled', False) %}
+        coalesce(student_assessments.school_year, dates_xwalk.school_year, {{derive_school_year('administration_date')}}) as school_year,
```
rlittle08, line 39: just in case, make this explicitly `student_assessments.administration_date`

rlittle08 (commented):
wrote some comments that mostly align with your list of potential drawbacks!

I also did a performance comparison on TX and this doesn't seem to add any time to execution there

models/core_warehouse/fct_student_assessment.sql
jalvord1: Yeah that would _require_ that the xwalk be added to the project in order to upgrade which doesn't feel ideal, but one could argue that's a better scenario than adding a bunch of vars

models/core_warehouse/fct_student_assessment.sql
jalvord1: Yeah wondering if there's a better method than creating a test? EX: we could write the join differently/more explicitely? Feels like either way the xwalk would need to be filled out in the correct way to match the logic so maybe we can't avoid there being a test

jalvord1:
separate question @rlittle08: do you think we should add capability to _override_ school years as part of this feature?

models/core_warehouse/fct_student_assessment.sql
rlittle08: ah great point, and we wanted this in a non-breaking release

rlittle08: I think we could keep as you have it and decide to change this behavior later, especially if we start to get feedback from others

models/core_warehouse/fct_student_assessment.sql
jalvord1: Sorry yeah I wasn't clear, we could remove from `edu_wh`

models/core_warehouse/fct_student_assessment.sql
rlittle08: the only more strict thing I can think of is have the macro throw a hard error if the xwalk doesn't meet expectations

rlittle08:
> separate question @rlittle08: do you think we should add capability to _override_ school years as part of this feature?

I do think that would be a positive change. do you know of any current cases for that?

jalvord1:
> > separate question @rlittle08: do you think we should add capability to _override_ school years as part of this feature?
> 
> I do think that would be a positive addition. do you know of any current cases for that?

Not any current asks for this that I know of, but I could see it coming up. Maybe not worth extra effort and could be a future addition? I'll think a little bit more on it.

models/core_warehouse/fct_student_assessment.sql
jalvord1: The macro is separate from the xwalk. Maybe it doesn't need to be that strict, so a test + directions for filling out the xwalk would be acceptable. Is it correct that Luda was assigned to work on a [seed properties file](https://docs.getdbt.com/reference/seed-configs)? That feels like a good place to add notes like this

models/core_warehouse/fct_student_assessment.sql
jalvord1: Keep truly as-is or make the variable name specific to assessments? I think I lean towards the latter

models/core_warehouse/fct_student_assessment.sql
rlittle08: yeah sounds good to me!

models/core_warehouse/fct_student_assessment.sql
rlittle08: yeah agreed

rlittle08:
Tests I did using TX:

Testing xwalk var:
- with xwalk not defined and `'edu:school_year:assessment_dates_xwalk_enabled'` not set. compiles and runs correctly
- with xwalk not defined and `'edu:school_year:assessment_dates_xwalk_enabled'` set to False. compiles and runs correctly
- with xwalk not defined and `'edu:school_year:assessment_dates_xwalk_enabled'` set to True. **compilation error. I think this is preferred behavior**

Testing macro with xwalk var disabled:
- with `edu:school_year:start_month` & start_day not defined: compiles and runs correctly (coalesce to null)
- with `edu:school_year:start_month` defined but `edu:school_year:start_day` not defined: **compiles with coalesce to null I think preferred behavior, but kind of quiet. Someone may think they've configured it correctly when they haven't**
- with `edu:school_year:start_month` & start_day both defined (08 and 01): compiles and runs correctly
- with `edu:school_year:start_month` & start_day both defined ('August' and '01'): **database error. maybe fine, could consider renaming variable to start_month_index**
- with `edu:school_year:start_month` & start_day explicitly set to `null`: correctly compiles with coalesce to null

Testing macro with xwalk var enabled:
- with xwalk created but empty (no rows): **database error: trying to cast a null from empty xwalk to date. how can we fix this?**
- with xwalk populated but no column config in dbt project: compiles and runs correctly
- with xwalk populated but dates left blank in seed e.g. `,,,`: compiles and runs correctly (doesn't join to anything)
- with xwalk populated with real dates and override_existing = False: compiles and runs correctly (only overrides nulls, and overrides them correctly within windows)
- with xwalk populated with real dates and override_existing = True, and assess ID/namespace left blank: compiles and runs correctly (overrides school year based on dates)
- with xwalk populated with real dates and override_existing = True, and assess ID/namespace populated: compiles and runs correctly (overrides school year only for assess configured). one note, this could turn into a ton of rows, for cases like CLI CIRCLE in TX, which has >30 assess ids

rlittle08:
I also added a test for overlapping dates, tested that with various error cases and is working

models/core_warehouse/fct_student_assessment.sql
rlittle08: test implemented

rlittle08 (approved):
talked over with Erik, and based on testing, good to merge

PR #119, opened 2024-03-24
make stu display name sql configurable

rlittle08:
This adds ability to configure the SQL logic for `stu_display_name` in `dim_student`, but keeps the existing logic by default.

Existing logic: `Lastname, Firstname, MiddleInitial`

Example custom logic: `Lastname, Firstname, Middle Suffix`

Is configuring this necessary, or should we just include generation_code_suffix for all?


rlittle08:
tested with and without variable defined ✔️

ejoranlienea (approved):

PR #126, opened 2024-04-24
Feature: Add `school_year` to `fct_student_grades`

johncmerfeld:
### Description
Simple as the title, really!

### Testting
Tested in a project implementation and the new column is present, populated with the school year values present in `stg_ef3__grades`

### DIFF
--- a/models/core_warehouse/fct_student_grades.sql
+++ b/models/core_warehouse/fct_student_grades.sql
@@ -34,6 +34,7 @@ formatted as (
         dim_course_section.k_course_section,
         dim_school.k_school,
         dim_grading_period.k_grading_period,
+        stg_grades.school_year,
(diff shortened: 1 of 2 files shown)


rlittle08 (approved):
ready for next edu_wh release

rlittle08:
One note on this for when we release - while it doesn't seem to be breaking, there is some possibility of downstream applications breaking, if they merge in school_year from other sources and don't use explicit namespaced reference. e.g. this kind of query:

```
select school_year, grade
from grades
join students on k_student
```

would error.

We saw this when we added school_year to other models in the past.

ejoranlienea:
Interesting. This is kind of already there through the `k_grading_period` key, which folks would tend to join when using this anyway, no? Was there a motivating case for this?

PR #127, opened 2024-04-26
Add PR template

rlittle08:


.github/pull_request_template.md
```diff
@@ -0,0 +1,69 @@
...
+## Tests and QC done:
+<!---
+Describe any process that confirms that the files do what is expected, include screenshots if relevant. For example:
+- Analyst replication confirmed that updates to `stg_model` new counts were correct.
+- Executed a dbt project run and ensured it was successful.
+-->
+
+## edu_wh PR Review Checklist:
+Make sure the following have been completed before approving this PR:
+- [ ] Description of changes has been added to Unreleased section of [CHANGELOG.md](/CHANGELOG.md). Add under `## New Features` for features, etc.
```
jalvord1, line 54: my only thought here is that because some of these options are either/or, it will say not all items are completed. Maybe that doesn't matter but just flagging

susanxiong: same goes for the high/med/low priority one. it's kind of annoying to look at but doesn't actually block merging i think.

rlittle08: many people want this behavior to change, hah https://github.com/orgs/community/discussions/5197

rlittle08:
@ejoranlienea can we merge this one with today's release?

ejoranlienea (approved):

PR #131, opened 2024-05-03
add earned_additional_credits in fct_course_transcript

keenzarate213:
## Description & motivation
based on this Monday [item](https://educationanalytics.monday.com/boards/3556827529/pulses/6532381985). Adding in additional credits in fct_course_transript keeping as variant

adding `earned_additional_credits` variant

## Tests and QC done:
Tested in `stadium_boston`

## PR Merge Priority:
- [x] Low
- [ ] Medium
- [ ] High

### DIFF
--- a/models/core_warehouse/fct_course_transcripts.sql
+++ b/models/core_warehouse/fct_course_transcripts.sql
@@ -42,15 +42,13 @@ formatted as (
         course_transcripts.method_credit_earned,
         course_transcripts.earned_credit_type,
         course_transcripts.earned_credit_conversion,
-        {{ edu_edfi_source.extract_descriptor('value:additionalCreditTypeDescriptor::string') }} as earned_additional_credit_type,
-        value:credits as earned_additional_credit,
         course_transcripts.attempted_credit_type,
         course_transcripts.attempted_credit_conversion,
         course_transcripts.assigning_organization_identification_code,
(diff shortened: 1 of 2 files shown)


ejoranlienea:
I don't know that we can change the grain of course transcript to get this in.
If we want this in a single table we might need to pivot it wide, such that (based on the screenshot) we'd get `other_earned_additional_credits`, or something along those lines.

rlittle08:
> I don't know that we can change the grain of course transcript to get this in. If we want this in a single table we might need to pivot it wide, such that (based on the screenshot) we'd get `other_earned_additional_credits`, or something along those lines.

@keenzarate213 can you remove the lateral flatten and keep it as a variant column for now?

rlittle08 (approved):
This looks good and fulfills a partner's request to expose the variant column in warehouse. 

There's some potential follow-up work here to extract the additional credits using a) a build model  with a pivot or b) Snowflake's [TRANSFORM](https://docs.snowflake.com/en/sql-reference/functions/transform).

However, both a) and b) have downsides
a) could be a costly model (transcripts are huge, so unpacking them into a new model is expensive), and in general we've run into issues pivoting models that will have unknown descriptor values
b) is tricky because the transformation we want here is actually a dbt macro (extract_descriptor)

PR #132, opened 2024-05-16
feature/fct_table_extensions

rlittle08:
## Description & motivation
Brings extensions through to all fct tables that directly reference a stg table. This was already done for a few tables (mostly program associations), but this PR adds to all.

Edge case: if an implementation has configured an extension with a colname that already exists in the fct table (e.g., the fct table joins multiple sources, and this column name exists in another source), there will be a duplicate column name error. This should be very rare, and easy to catch in testing.

## PR Merge Priority:
- [ ] Low
- [x] Medium
- [ ] High

## Changes to existing files:
- several fct models: add call to `extract_extension()`, which pulls out any extension fields that have been configured in `vars:extensions`

## New files created:
None

## Tests and QC done:
Tested in Texas, where there are many extensions configured. Testing included models without extension, where I confirmed macro compiles to nothing. 


johncmerfeld (commented):
Looks good to me. I grepped for other files that directly reference staging and found the following files not included here:
```bash
grep --exclude="dim*" -rni "select \* from {{ ref('stg_ef3__" .
```
  - fct_student_discipline_incident_summary.sql
  - fct_student_discipline_actions_summary.sql

Upon looking at these models it makes sense to me why they're not included; their grain seems different from the other fact tables. But I wanted to highlight this before approving just to confirm they were intentionally excluded from this

rlittle08:
> Looks good to me. I grepped for other files that directly reference staging and found the following files not included here:
> 
> ```shell
> grep --exclude="dim*" -rni "select \* from {{ ref('stg_ef3__" .
> ```
> 
> * fct_student_discipline_incident_summary.sql
> * fct_student_discipline_actions_summary.sql
> 
> Upon looking at these models it makes sense to me why they're not included; their grain seems different from the other fact tables. But I wanted to highlight this before approving just to confirm they were intentionally excluded from this

Yeah, we probably _could_ add extensions here, but I guess I'm not sure if it makes sense, since they both are pulling together from multiple stg tables and that could get confusing

johncmerfeld:
> Yeah, we probably _could_ add extensions here, but I guess I'm not sure if it makes sense, since they both are pulling together from multiple stg tables and that could get confusing

That's fair. Although there are a couple existing fct tables that pull in multiple sets of extensions, e.g. models/core_warehouse/fct_staff_school_association.sql

rlittle08:
> > Yeah, we probably _could_ add extensions here, but I guess I'm not sure if it makes sense, since they both are pulling together from multiple stg tables and that could get confusing
> 
> That's fair. Although there are a couple existing fct tables that pull in multiple sets of extensions, e.g. models/core_warehouse/fct_staff_school_association.sql

good point - can you draft an update to these disc models with the extensions?

rlittle08:
> > > Yeah, we probably _could_ add extensions here, but I guess I'm not sure if it makes sense, since they both are pulling together from multiple stg tables and that could get confusing
> > 
> > 
> > That's fair. Although there are a couple existing fct tables that pull in multiple sets of extensions, e.g. models/core_warehouse/fct_staff_school_association.sql
> 
> good point - can you draft an update to these disc models with the extensions?

@johncmerfeld that looks good, but i just realized, the extensions are not pulled out in the flattened stg tables e.g. https://github.com/edanalytics/edu_edfi_source/blob/main/models/staging/edfi_3/stage/stg_ef3__discipline_actions__student_discipline_incident_behaviors.sql so we could either add an edu_edfi_source feature branch or leave this for another day

johncmerfeld:
> @johncmerfeld that looks good, but i just realized, the extensions are not pulled out in the flattened stg tables e.g. https://github.com/edanalytics/edu_edfi_source/blob/main/models/staging/edfi_3/stage/stg_ef3__discipline_actions__student_discipline_incident_behaviors.sql so we could either add an edu_edfi_source feature branch or leave this for another day

Eh, I've got time, I'll open a PR in edfi source!

johncmerfeld:
@rlittle08 it's up

johncmerfeld (approved):

PR #143, opened 2024-07-29
Update to dbt_utils 1.0+ (1.3.0)

gnguyen87:
## 

## Description & motivation

Update dbt_utils  package to the latest version.

Breaking changes exist in the new update with regards to `surrogate.key()`. The original treats nulls and empty strings the same. The new one does not. There's a variable that can be added to still opt into this original treatment. 

**Dependent on https://github.com/edanalytics/edu_edfi_source/pull/97 and https://github.com/edanalytics/edu_ext_podium/pull/54 **

## PR Merge Priority:

- [x] Low 
- [ ] Medium
- [ ] High

## Changes to existing files:

- Any `dbt_utils.surrogate_key()` is replaced with `dbt_utils.generate_surrogate_key()`
- Add a var in `dbt_project.yml` to opt in to [legacy behavior of this function](https://github.com/dbt-labs/dbt-utils/releases#:~:text=surrogate_key()%20has%20been%20replaced%20by%20generate_surrogate_key().%20You,be%20regenerated%20with%20new%20values%20without%20issue.)

## Tests and QC done
- Install locally with a separate dbt profile and all models ran.


rlittle08:
one note, this should be saved for next "major" point release (0.4.0), because it has potential to break implementations who use `surrogate_key` in stadium-specific models, so it will require an extra migration

PR #144, opened 2024-07-29
fix stu-school-att unique test

rlittle08:


sleblanc23 (commented):
I think the primary key documentation in line 13 should be updated too

rlittle08:
> I think the primary key documentation in line 13 should be updated too

done!

sleblanc23 (approved):

PR #155, opened 2024-10-03
use calendars rather than enrollments for tenant lea ownership

ejoranlienea:
## Description & motivation
In rare cases, we need to identify which LEAs are actually members of a tenant. While we were doing this via enrollment records, these can be unpopulated at the start of the school year, while calendars should be populated immediately.

## PR Merge Priority:

- [ ] Low
- [x] Medium
- [ ] High

## Changes to existing files:
Switch the base table for tenant-lea attribution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 ## New features
 ## Under the hood
+- Tweak tenant-lea attribution logic
 ## Fixes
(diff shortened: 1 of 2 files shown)


rlittle08:
This appears to work, and in TX, it adds one new tenant (who has calendars but not enrollments populated). I think we go ahead and merge this

PR #158, opened 2024-12-06
Fix/att event multi calendars

ryanaguilar:
## Description & motivation

This PR addresses an edge case that occured in the BPS Stadium implementation where a student had multiple enrollments associated with different calendars per enrollment, which resulted in a student's attendance event not contributing to their attendance count.

Currently, the attendance_event_date is converted into a k_calendar_date by joining to fct_student_school_association and deduping by selecting enrollments where is_latest_annual_entry is true.

This updated model would instead join so that the attendance_event_date is within the range of the enrollment's entry_date and exit_withdraw_date (and accounts for null exit_withdraw_dates). 

However, if a student has overlapping enrollments at the same school, multiple rows will
be returned for each date, and so an additional deduplication step is introduced.

See the original Monday item [here](https://educationanalytics.monday.com/boards/2610522828/pulses/7737880198).

No breaking changes, but in rare cases could change attendance count.

## PR Merge Priority:

- [x] Medium

## Future ToDos & Questions:
- Further discussion may be needed on how to pick a winner when there overlapping enrollments with multiple calendars at the same school. The rule implemented here is one option that was presented during planning.

### DIFF
--- a/models/core_warehouse/fct_student_school_attendance_event.sql
+++ b/models/core_warehouse/fct_student_school_attendance_event.sql
@@ -45,7 +45,6 @@ fct_student_school_assoc as (
     the same calendar to be sure.
     */
     select * from {{ ref('fct_student_school_association') }}
-    where is_latest_annual_entry
 ),
 xwalk_att_events as (
     select * from {{ ref('xwalk_attendance_events') }}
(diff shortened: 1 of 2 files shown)


models/core_warehouse/fct_student_school_attendance_event.sql
```diff
@@ -79,9 +72,40 @@ formatted as (
         on stg_stu_sch_attend.k_student = fct_student_school_assoc.k_student
         and stg_stu_sch_attend.k_school = fct_student_school_assoc.k_school
     join dim_calendar_date
-        on fct_student_school_assoc.k_school_calendar = dim_calendar_date.k_school_calendar
-        and stg_stu_sch_attend.attendance_event_date = dim_calendar_date.calendar_date
+         on fct_student_school_assoc.k_school_calendar = dim_calendar_date.k_school_calendar
+         and stg_stu_sch_attend.attendance_event_date = dim_calendar_date.calendar_date
+         and dim_calendar_date.calendar_date between fct_student_school_assoc.entry_date and coalesce(fct_student_school_assoc.exit_withdraw_date,getdate())
```
rlittle08, line 77: any specific to use getdate() (returns a timestamp) instead of current_date() (returns a datestamp) here?

ryanaguilar: No specific reason, I don't think I realized there was a difference.

rlittle08: I don't think it affects the logic in this case, but I prefer current_date(), since the extra precision isn't needed and current_date() is available in databricks

rlittle08 (approved):

rlittle08:
I tested and deeply qc'd in Boston, TX, SC. Only changes were expected improvements. LGTM

PR #168, opened 2025-04-16
Feature/Databricks support

sleblanc23:
## Description & motivation
This PR, along with [these changes in `edu_edfi_source`](https://github.com/edanalytics/edu_edfi_source/pull/129), makes EDU compatible with Databricks.

There are several changes that might qualify as breaking, but don't actually require any migration. 
- All columns which are part of the primary key of a table are set explicitly as not null
- The primary key of `fct_student_gpa` has been changed to remove `is_cumulative`, but the logic was adjusted so that the effective grain is the same
- Added `fct_staff_school_association.k_staff_school_association` and updated the primary key
- Changed column order in `fct_student_assessment` and ``fct_student_objective_assessment`
- Changed `dim_class_period.start_time` and `end_time` from time data types to strings

## Changes to existing files:
- Databricks requires that every column which is part of the primary key must be set as not null and enforces this constraint. This was only a challenge in two models:
  - In `fct_student_gpa`, `is_cumulative` was part of the primary key due to changes in the model. Refactored to remove it from the primary key and instead ensure that `gpa_type` accounts for the deprecated behavior
  - In `fct_staff_school_association`, `staff_classification` and `begin_date` are sometimes null due to how we left join data from `staffEdOrg` onto the `staffSchool` data. Created a surrogate key to get around this issue
- Most other changes are a straightforward substitution of a macro call in place of Snowflake-specific array behavior or to account for syntax differences. See the documentation of these macros in `edu_edfi_source` for more details.  
- `fct_student_assessment` was rearranged slightly to add v_other_results on at the end because variant columns can't be included in a group by in Databricks
- `fct_student_discipline_actions` was updated with slightly simpler logic to select one `k_staff` for the singular field in case there are multiple. We use max() to consistently select one instead of the first one in the array. This avoids reliance on `index`, which is called `pos` in Databricks.
- `dim_class_period` was updated to handle the conversion to 24 hour time with string manipulation and the period duration calculation via the timestamp type instead of time type, which Databricks doesn't have.
- `bld_ef3__course_char__combined_long` was refactored to avoid array concatenation, which works differently in Databricks. The rewrite does not slow down the model much (adds ~1 sec in SC).
- Small changes in a few custom tests to use syntax supported by both platforms

## Tests and QC done:
- Ran in Databricks on demo data; confirmed that all models ran and did some light checking on contents
- Ran in CA and SC; confirmed that affected models are unchanged by comparing row counts and relevant column values

### DIFF
--- a/models/build/edfi_3/courses/bld_ef3__course_char__combined_long.sql
+++ b/models/build/edfi_3/courses/bld_ef3__course_char__combined_long.sql
@@ -1,53 +1,70 @@
 -- this model unpacks the characteristics payload from each of the 3 levels at 
--- which it's defined, then combines the lists across the 3 levels before
--- unpacking it back to section by characteristic grain
-with section_chars as (
-    select  
-        k_course_offering,
-        k_course_section,
-        array_agg({{ edu_edfi_source.extract_descriptor('value:sectionCharacteristicDescriptor::string') }}) as section_characteristic
+-- which it's defined, then unions the lists at the section-characteristic grain
+with sections as (
+    select *
     from {{ ref('stg_ef3__sections') }}
-        {{ json_flatten('v_section_characteristics', outer=True) }}
-    group by 1,2
+),
+offerings as (
+    select *
+    from {{ ref('stg_ef3__course_offerings') }}
+),
+courses as (
+    select *
+    from {{ ref('stg_ef3__courses') }}
+),
+section_chars as (
+    select
+        sections.tenant_code, 
+        sections.api_year,
+        offerings.k_course,
+        sections.k_course_offering,
+        sections.k_course_section,
+        {{ edu_edfi_source.extract_descriptor('value:sectionCharacteristicDescriptor::string') }} as characteristic
+    from sections
+    join offerings
+        on sections.k_course_offering = offerings.k_course_offering
+        {{ edu_edfi_source.json_flatten('sections.v_section_characteristics', outer=True) }}
 ), 
 offering_chars as (
     select 
-        k_course,
-        k_course_offering,
-        array_agg({{ edu_edfi_source.extract_descriptor('value:courseLevelCharacteristicDescriptor::string') }}) as offering_characteristic
-    from {{ ref('stg_ef3__course_offerings') }}
-        {{ json_flatten('v_course_level_characteristics', outer=True) }}
-    group by 1,2
+        offerings.tenant_code, 
+        offerings.api_year,
+        offerings.k_course,
+        offerings.k_course_offering,
+        sections.k_course_section,
+        {{ edu_edfi_source.extract_descriptor('value:courseLevelCharacteristicDescriptor::string') }} as characteristic
+    from offerings
+    join sections
+        on sections.k_course_offering = offerings.k_course_offering
+        {{ edu_edfi_source.json_flatten('offerings.v_course_level_characteristics', outer=True) }}
 ),
 course_chars as (
     select 
-        tenant_code, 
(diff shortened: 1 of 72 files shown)


rlittle08:
@sleblanc23 can we do a deeper QC check on some of the affected models to be extra safe about value changes (thinking gpa, staff, assessment)? If helpful, we could bring in another DE for that

alchenist:
> @sleblanc23 can we do a deeper QC check on some of the affected models to be extra safe about value changes (thinking gpa, staff, assessment)? If helpful, we could bring in another DE for that



Audited data in Boston for the following higher-priority models:


| model                            | row count    | value check (with hashes)                                                                                                                                              | notes                                                                                                  |
| -------------------------------- | ------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
| fct_student_gpa                  | ✅ equivalent | confirmed that changes are limited to `gpa_type` ('Unknown' -> \['Cumulative, unknown weighting', 'Non-cumulative, unknown weighting'\]);                              | Boston doesn't have populated GPAs in the pre-DS5 optional location on student academic records anyway |
| fct_staff_school_association     | ✅ equivalent | ✅ equivalent across `k_staff, k_school, school_year, program_assignment, staff_classification, begin_date, tenant_code`                                                |                                                                                                        |
| fct_student_assessment           | ✅ equivalent | ✅ equivalent across `k_student_assessment, k_assessment, k_student, tenant_code, student_assessment_identifier, v_other_results`                                       |                                                                                                        |
| fct_student_objective_assessment | ✅ equivalent | ✅ equivalent across `k_student_objective_assessment, k_objective_assessment, k_student_assessment, k_assessment, k_student, tenant_code, school_year, v_other_results` |                                                                                                        |



Value checks used the specified columns and were based off of [`audit_helper.compare_and_classify_query_results`](https://github.com/dbt-labs/dbt-audit-helper?tab=readme-ov-file#compare_and_classify_query_results-source).

models/build/edfi_3/courses/bld_ef3__course_char__combined_long.sql
```diff
@@ -1,53 +1,70 @@
...
-    group by 1,2
+        offerings.tenant_code, 
+        offerings.api_year,
+        offerings.k_course,
+        offerings.k_course_offering,
+        sections.k_course_section,
+        {{ edu_edfi_source.extract_descriptor('value:courseLevelCharacteristicDescriptor::string') }} as characteristic
+    from offerings
+    join sections
+        on sections.k_course_offering = offerings.k_course_offering
```
rlittle08, line 38: i'm being picky - [please specify 'from' table first ](https://edanalytics.slite.com/app/docs/Uj83Z7k81B#dc420ba2)

models/build/edfi_3/courses/bld_ef3__course_char__combined_long.sql
```diff
@@ -1,53 +1,70 @@
...
-    join section_chars 
-        on offering_chars.k_course_offering = section_chars.k_course_offering
+unioned as (
+    select * from section_chars
+    union all
+    select * from offering_chars
+    union all
+    select * from course_chars
 )
 select
```
rlittle08, line 63: i think we need a distinct/group by here to replicate the previous 'array_distinct' behavior

rlittle08 (changes requested):
looked through everything and all checks out, just a couple requests on the bld course chars model!

rlittle08 (approved):

PR #169, opened 2025-05-22
update key to include begin_date, ds 5.0

ejoranlienea:
## Description & motivation
Add `begin_date` to primary key to align with DS 5.0+. Only changing the primary key declaration and corresponding docs.
Note that `begin_date` should be guaranteed non-null, as even before it was required it had a default value, so there should be no impact on Databricks compatibility.

## PR Merge Priority:

- [ ] Low
- [x] Medium
- [ ] High

-->

### DIFF
--- a/models/core_warehouse/fct_staff_section_association.sql
+++ b/models/core_warehouse/fct_staff_section_association.sql
@@ -1,7 +1,7 @@
 {{
   config(
     post_hook=[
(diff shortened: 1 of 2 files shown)


rlittle08:
@ejoranlienea I was gonna say we need to bump the packages dependency, but I guess maybe not, since we're just making the key 'more strict' here?

PR #172, opened 2025-06-13
Fix/Support multiple subjects per course in DS 5.0

alchenist:
## Description & motivation
Handle DS 5.0+, where [academic subjects are now optional collections for Courses](https://edfidocs.blob.core.windows.net/$web/handbook/v5.0/index.html#/Course237aca05-9476-4646-9640-ecb9630bd8c7). `academicSubject` was previously [single-valued](https://edfidocs.blob.core.windows.net/$web/handbook/v4.0/index.html#/Course550). 

Depends on edanalytics/edu_edfi_source#139

- New array-valued column in `dim_course` may affect downstream queries
- `dim_course.academic_subject` will now be populated for DS 5.0 data; previously null

## PR Merge Priority:
- [x] Low
- [ ] Medium
- [ ] High

## Changes to existing files:

- `dim_course`: 
  - Update single-valued `academic_subject` as follows:
    - For courses in DS 5.0+, take the first value if the array has only one value. Otherwise, yield 'Multiple.`
    - For courses prior to DS 5.0, maintain old behavior and take the original value
  - Add new array-valued `subject_array` column, containing the full array of academic subjects for any DS 5.0+ rows.
  - Note that the above changes do not currently depend on `data_model_version` and merely depend on the presence or absence of the new array-valued field in the source data.

## New files created:
- `bld_ef3__course_subject`: Contains one row per course and wraps multiple subjects per course back into an array; also contains coalesce logic to produce a single value in `academic_subject` from whatever source data is present, single- or array-valued.

## Tests and QC done:
- Ran against dev data in Boston, which has sideloaded course data from DS 5.0. Confirmed row counts stay the same for `dim_course` and that new logic works properly for DS 5.0+ data. Unable to confirm if the multiple-valued logic works; no source data with more than one subject per course yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 ## Under the hood
 ## Fixes
 
+
+# edu_wh v0.4.3
+## Fixes
+- Fix model `dim_course` to handle multiple academic subjects per course (Ed-Fi Data Standard v5.0 breaking change)
+  - Add array column `subject_array` to `dim_course`, containing array of academic subjects if these exist
+  - Add logic to populate `academic_subject` column with single-valued subjects in both cases where data source is <5.0 or >5.0
+  - Add upstream `bld_ef3__course_subject` 
+
 # edu_wh v0.4.2
 ## New features
 - Add tests `cfg_assessment_scores` and `cfg_objective_assessment_scores` to find assess/obj assess with no scores configured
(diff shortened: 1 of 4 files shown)


alchenist:
> code all looks good. The last thing we should do before merging is let the team know about the choice for 'Multiple' instead of picking one. You could bring this to our next Thursday 6/26 product dev meeting, or post in #edu_stadium_engineering

Great, I can do a quick update on 6/26!

rlittle08 (approved):

PR #179, opened 2025-07-17
Staff Email Handling

ejoranlienea:
## Description & motivation
This adds two primary features:

First, Ed-Fi introduced `StaffEducationOrganizationContactAssociation`, which provides a second possible place to put email addresses (in addition to `staffs` itself). This PR adds a build model that combines emails from the two sources into one unified list
.
Second, it also adds a second build model on top of this that attempts to filter this down to only official/work emails, which can be useful for syncing to authentication/authorization systems, or other use-cases that require only valid, official email addresses. It does this by applying 3 filters:

1. Check whether the email address is structurally valid
2. Filter to email addresses tagged with a descriptor that indicates official/work email (configurable)
3. In case of data entry errors, also attempt to remove emails from personal-looking domains (also configurable). 

## PR Merge Priority:

- [ ] Low
- [ ] Medium
- [ ] High

### DIFF
--- a/models/build/edfi_3/staffs/bld_ef3__staff_emails.sql
+++ b/models/build/edfi_3/staffs/bld_ef3__staff_emails.sql
@@ -1,0 +1,28 @@
+-- combine staff emails from multiple sources in ed-fi
+with staff_email as (
+    select
+        tenant_code,
+        k_staff,
+        null as k_lea,
+        null as k_school,
+        email_type,
+        email_address
+    from {{ ref('stg_ef3__staffs__emails') }}
+),
+staff_edorg_email as (
+    select
+        seoca.tenant_code,
+        seoca.k_staff,
+        seoca.k_lea,
+        seoca.k_school,
+        seoca.contact_title as email_type,
+        seoca.email_address
+    from {{ ref('stg_ef3__staff_education_organization_contact_associations') }} as seoca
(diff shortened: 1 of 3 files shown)


models/build/edfi_3/staffs/bld_ef3__staff_emails.sql
```diff
@@ -0,0 +1,37 @@
...
+    select * from staff_email
+    union all
+    select * from staff_edorg_email
+)
+select
+    *,
+    -- allow dots, hyphens, and underscores in email (and optionally plus-addressing)
+    -- but don't allow apostrophes, spaces, other characters
+    -- allow final URL component to be between 2 and 9 characters
+    email_address rlike '[a-zA-Z0-9.-_]+[+]?[a-zA-Z0-9.-]*@[a-zA-Z0-9.-]+[.][a-zA-Z0-9]{2,9}' as is_valid_email
```
rlittle08, line 36: do we want to use `^` and `$` to ensure full string match instead of partial?

ejoranlienea: makes sense. 
while testing this, I realized another small bug I had introduced. `.-_` gets interpreted as 'the range between . and _' in some parsers. You can avoid this by escaping the `-`, but escape semantics can be tricky across parsers. Reordering as `_.-` appears to get the intended result of including all three characters literally in the legal character set.

PR #231, opened 2026-04-10
feature/:Enhanced dim_course_section to expose section_characteristics descriptors

smckee-tnedu:
## Description & motivation
After the fix for course_level_characteristics, the visibility of section characteristics was gone. This set of changes brings that visibility back.

This change requires a new xwalk: `xwalk_section_characteristics.csv`. An example looks like this:
```
section_characteristic_descriptor,indicator_name
Fall Block,is_fall_block
Spring Block,is_spring_block
Year-long,is_year_long
``` 

## PR Merge Priority:
- [X] Low
- [ ] Medium
- [ ] High

## Changes to existing files:
- `models/core_warehouse/dim_course_section.sql` : Uses a new build model to expose the descriptors that come in on the sections_characteristics field.

## New files created:
- `models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql` : Builds a set of wide columns that can be accordioned in dim_course_section.

## Tests and QC done:
This has been tested within stadium_tennessee.

- [ ] Code has been tested/checked for Databricks and Snowflake compatibility - EA engineers see Databricks checklist [here](https://edanalytics.slite.com/app/docs/LRjXFxVRAWA5ST) 

### DIFF
--- a/models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql
+++ b/models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql
@@ -1,0 +1,35 @@
+with sections as (
+    select * from {{ ref('stg_ef3__sections') }}
+),
+xwalk_section_characteristics as (
+    select * from {{ ref('xwalk_section_characteristics') }}
+),
+flattened as (
+    select 
+        k_course_section,
+        {{ edu_edfi_source.extract_descriptor('section_chars.value:sectionCharacteristicDescriptor::string') }} as section_characteristic
+    from sections
+        {{ edu_edfi_source.json_flatten('v_section_characteristics', 'section_chars', outer=true) }}
+),
+pivoted as (
+    select 
+        k_course_section,
+        {{ edu_edfi_source.json_array_agg(
+            'section_characteristic',
+            order_by='section_characteristic',
+            is_terminal=True
+        ) }} as section_characteristics_array
+        {%- if not is_empty_model('xwalk_section_characteristics') -%},
+            {{ ea_pivot(
+                    column='indicator_name',
+                    values=dbt_utils.get_column_values(ref('xwalk_section_characteristics'), 'indicator_name'),
+                    cast='boolean',
+            ) }}
+        {%- endif %}
+    from flattened
+    left outer join xwalk_section_characteristics 
+        on flattened.section_characteristic = xwalk_section_characteristics.section_characteristic_descriptor
+    group by all
+)
+select *
+from pivoted
--- a/models/core_warehouse/dim_course_section.sql
+++ b/models/core_warehouse/dim_course_section.sql
@@ -22,8 +22,11 @@ stg_ef3__sections as (
 dim_course as (
     select * from {{ ref('dim_course') }}
 ),
-section_chars as (
+course_level_chars as (
     select * from {{ ref('bld_ef3__course_char__combined_wide') }}
+),
+section_chars as (
+    select * from {{ ref('bld_ef3__section__wide_section_characteristics') }}
 ),
 joined as (
     select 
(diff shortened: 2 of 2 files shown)


models/core_warehouse/dim_course_section.sql
AngelicaLastra, line 80: Since we are adding them now through this PR, we no longer need the stale todo. We can remove this comment to ensure we don't cause confusion in the future.

models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql
```diff
@@ -0,0 +1,35 @@
...
+        ) }} as section_characteristics_array
+        {%- if not is_empty_model('xwalk_section_characteristics') -%},
+            {{ ea_pivot(
+                    column='indicator_name',
+                    values=dbt_utils.get_column_values(ref('xwalk_section_characteristics'), 'indicator_name'),
+                    cast='boolean',
+            ) }}
+        {%- endif %}
+    from flattened
+    left outer join xwalk_section_characteristics
```
AngelicaLastra, line 30: Very small, picky thing, can we use `left join` instead? It's the standard for the repo even if it means the same thing.

models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql
```diff
@@ -0,0 +1,35 @@
...
+),
+xwalk_section_characteristics as (
+    select * from {{ ref('xwalk_section_characteristics') }}
+),
+flattened as (
+    select 
+        k_course_section,
+        {{ edu_edfi_source.extract_descriptor('section_chars.value:sectionCharacteristicDescriptor::string') }} as section_characteristic
+    from sections
+        {{ edu_edfi_source.json_flatten('v_section_characteristics', 'section_chars', outer=true) }}
```
AngelicaLastra, line 12: I wonder if we need the `outer=true` here? What would be the benefit of creating rows for course sections without characteristics for this model?

AngelicaLastra (changes requested):
@smckee-tnedu  this is a really clean PR! Thank you for catching this gap in the `dim_course_section` model!

There are a few small things to clean up before merge, but they are very small:
- In `bld_ef3__section__wide_section_characteristics`:
  - `left outer join` should be `left join` to match every other model in the repo. 
  - I am wondering if we need the `outer=True`? what is the benefit of pushing through rows without a section characteristic?

- In `dim_course_section`:
  - The comment `-- todo: add characteristic indicators` on line 80 is stale, we should remove it so it doesn't cause confusion in the future.

On the YML side, three things need to be added:
- `bld_ef3__section__wide_section_characteristics` is not registered in `_edfi_3__courses.yml`. It needs an entry with `config: tags: ['core']` and a column entry for `section_characteristics_array` to match the pattern established by `bld_ef3__course_char__combined_wide`. 
- In `dim_course_section.yml`, `section_characteristics_array` is missing a column entry (the `course_level_characteristics_array` entry is a good template to follow). The "Important Business Rules" block in the model description also only references `xwalk_course_level_characteristics` and the course-level `is_*` columns; it should be updated to mention `xwalk_section_characteristics` and the section-level indicator columns as well.

These are all very minor, final touch up changes. The code runs successfully, does what it intends to do, and does not negatively affect grain or other aspects of the model. Thank you again for this work!

models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql
smckee-tnedu: I don't think it's needed. I'ts been awhile since I wrote this but I might have been thinking that I saw outer=true as sort of a standard you guys were doing for these things? So that maybe if you had to troubleshoot data in the table, you'd have all the rows to look at and not just the rows that contained useful data? I don't know, to be honest.

models/build/edfi_3/courses/bld_ef3__section__wide_section_characteristics.sql
smckee-tnedu: You can, yes. They mean the same but I've always used the word "outer" just because in my history in IT (I'm old) I've always trained junior developers and explicitly saying "outer" has been helpful in reducing confusion.

models/core_warehouse/dim_course_section.sql
smckee-tnedu: Yep.

rlittle08:
marking for 0.7.0 because it requires xwalk update. very likely to release in ~August

PR #253, opened 2026-06-15
Feature/finance

rlittle08:
## Description & motivation
Adds a finance warehouse domain with dimension and fact models sourced from the EdFi finance staging layer introduced in the corresponding `edu_edfi_source` PR (`feature/finance_revamp`). Models are published under a dedicated `finance_warehouse` schema.

Depends on: `edu_edfi_source` `feature/finance_revamp` (packages.yml must be reverted to a pinned release before merge).

None, additive. To enable, will need to set var `` and configure folder to schema `prod_finance`

## PR Merge Priority:
- [ ] Low
- [x] Medium
- [ ] High

_Requires `edu_edfi_source` finance release to be cut and `packages.yml` reverted before this can merge._

Promised by end of June.

## Changes to existing files:
- `dbt_project.yml`: Added `finance_warehouse` schema configuration

## New files created:
- `models/build/edfi_3/bld_ef3__chart_of_accounts_dimensions.sql` / `.yml`: Build model that left-joins `stg_ef3__chart_of_accounts` to all 8 dimension staging tables, resolving human-readable codes and names for each dimension. Used by both `dim_chart_of_account` and `dim_local_account`.
- `models/finance_warehouse/dim_chart_of_account.sql` / `.yml`: Dimension table for chart of accounts, including resolved codes and names for all 8 dimension types.
- `models/finance_warehouse/dim_local_account.sql` / `.yml`: Dimension table for local accounts, joining to `dim_chart_of_account` for the chart-of-accounts attributes.
- `models/finance_warehouse/fct_local_actual_snapshots.sql` / `.yml`: Fact table of point-in-time actual financial snapshots, with `is_latest_snapshot` flag to identify the most recent record per account.
- `models/finance_warehouse/fct_local_budget_snapshots.sql` / `.yml`: Fact table of point-in-time budget snapshots, same structure as actuals.

## Tests and QC done:
- [x] dbt run completed successfully against finance models
- [x] `dim_chart_of_account` and `dim_local_account` row counts verified against source
- [x] `fct_local_actual_snapshots` / `fct_local_budget_snapshots` snapshot grain verified
- [x] `is_latest_snapshot` behavior confirmed correct

- [x] Code has been tested/checked for Databricks and Snowflake compatibility - EA engineers see Databricks checklist [here](https://edanalytics.slite.com/app/docs/LRjXFxVRAWA5ST) 


models/build/edfi_3/bld_ef3__chart_of_accounts_dimensions.sql
AngelicaLastra, line 1: It seems that all the `*_dimension_fiscal_year` columns are an exact copy of the chart of account's fiscal_year value. I queried the tables and got 100% match between the values for all rows. Given this, I think we should drop them from the build table, as they are only generating noise.

dbt_project.yml
```diff
@@ -29,6 +29,8 @@ models:
         +enabled: "{{ var('edu:tpdm:enabled', False) }}"
     core_warehouse:
       +schema: wh
+    finance_warehouse:
+      +schema: finance_wh
```
AngelicaLastra, line 33: do we want this in `wh` instead?

models/finance_warehouse/fct_local_actual_snapshots.yml
```diff
@@ -0,0 +1,42 @@
...
+    columns:
+      - name: k_local_account
+        description: Foreign key to dim_local_account
+        tests:
+          - not_null
+          - relationships:
+              to: ref('dim_local_account')
+              field: k_local_account
+      - name: k_local_actual_snapshot
+        description: Primary key — surrogate for the actual record at account / as-of / collection grain
```
AngelicaLastra: I dont think financial collection is a part of the grain, it is not a required field in edfi, and it comes as a string not a list of potential values. The grain here should be account for the tenant/fiscal year/as of

models/finance_warehouse/fct_local_budget_snapshots.yml
```diff
@@ -0,0 +1,42 @@
...
+    columns:
+      - name: k_local_account
+        description: Foreign key to dim_local_account
+        tests:
+          - not_null
+          - relationships:
+              to: ref('dim_local_account')
+              field: k_local_account
+      - name: k_local_budget_snapshot
+        description: Primary key — surrogate for the budget record at account / as-of / collection grain
```
AngelicaLastra: same as [this](https://github.com/edanalytics/edu_wh/pull/253/changes#r3444628890)

dbt_project.yml
AngelicaLastra: Do we need to add an enabled flag in the dbt project yaml like we do for tpdm? since these models are able to be disabled.

models/finance_warehouse/fct_local_actual_snapshots.yml
```diff
@@ -0,0 +1,42 @@
...
+        tests:
+          - unique
+          - not_null
+      - name: tenant_code
+        description: Data owner identifier
+      - name: api_year
```
AngelicaLastra: we are not using api year in this table

models/finance_warehouse/fct_local_budget_snapshots.yml
```diff
@@ -0,0 +1,42 @@
...
+        tests:
+          - unique
+          - not_null
+      - name: tenant_code
+        description: Data owner identifier
+      - name: api_year
```
AngelicaLastra: we are not using api year in this model

models/finance_warehouse/dim_chart_of_account.sql
AngelicaLastra, line 1: We can simplify this model a bit further by only selecting from the build model (instead of the stage + build combo). All of the columns we are selecting here are available in the build except for `v_ext`, which we can easily add.

On the k_*_dimension keys in the TODO block: I think we're fine to drop them from `dim_chart_of_account` while keeping them in the build model. Here's my reasoning, though realistically I could be swayed the other way:
1. They aren't foreign keys to anything in edu_wh, only to the staging models they belong to. There are no dim_ models for these keys, the keys only resolve against staging models, so there's nothing in the warehouse that's valuable to join to.
2. The information they point at is already flattened into the build model. We carry each of the staging model's `code` and `code_name`, so the keys don't unlock anything we don't already have on the row.
3. They add noise and make the table wider than it needs to be.

Also, per the comment in the build model, I think we should get rid of the `*_dimension_fiscal_year` columns, they are an exact match to the chart of accounts' `fiscal_year` value.

models/finance_warehouse/dim_local_account.sql
AngelicaLastra, line 1: I think we should get rid of the build model join here, we can get everything we need for this model just from the staging model.

All the values coming from the build are repetitive, since they already exist in the `dim_chart_of_accounts` model. We should only have to provide the `k_chart_of_account` key to join back to the other dim. That way this dim models stays purely for local interpretations.

AngelicaLastra (changes requested):
Wow! this was a big undertaking that made my head spiral a bit. Great job wrangling all of it! 

I think there are some places where we can trim logic out of some models to make them more intentional and easier to maintain.

I am also thinking we need to add a var for enabling/disabling these models given the `edu_edfi_source` configuration. Those should act together so we should probably use the same env var here. I also wonder if we want these to just land on the `wh` schema vs the `finance_wh` schema. If we wanted this in a finance specific schema maybe we should just have a `finance` one.

Lastly, I am wondering if a `fct_local_budget_to_actual_snapshots` (maybe with a shorter name) model would be good to have, where we join together `fct_local_actual_snapshots` and `fct_local_budget_snapshots` (full outer join?) and answer the question "For each account, how does what the org actually spent compare to what they budgeted, at any point in time?"

AngelicaLastra (approved):

PR #262, opened 2026-07-22
Add core warehouse models for Ed-Fi Survey Domain

gfitzgerald-ea:
## Summary

Initial core warehouse star schema for the Ed-Fi Survey Domain, per the [Q2.26 Survey Domain Models](https://edanalytics.slite.com/app/docs/-uBvjCAmjCIS8N) tech spec. Supports the SC School Climate Survey (Resonant) work.

**Dimensions**
- `dim_survey` — survey definitions and metadata
- `dim_survey_question` — question definitions
- `dim_survey_question_response_choice` — available response choices (kept as a separate entity; joins to response values only under specific circumstances, per the spec's "variable join key" note)

**Facts** (three grains, coarsest → finest)
- `fct_survey_response` — 1 submitted survey response
- `fct_survey_question_response` — 1 response to 1 question
- `fct_survey_question_response_value` — 1 selected/entered value per question response

## Design notes
- Respondent identity uses direct respondent keys on the facts (`k_student`, `k_staff`, `k_parent`) with a derived `respondent_type`, rather than a separate respondent dimension — matches the spec and avoids snowflaking. `respondent_type` is derived from the staging keys, not from join success.
- The spec references `dim_contact`; this package uses the existing `dim_parent` (`k_parent`) instead, for consistency with the rest of `edu_wh`. The staging `k_contact` column is aliased to `k_parent` in `fct_survey_response`.
- Multi-select delimited-text parsing (e.g. `Tutoring;Counseling;Mentoring` → one row per value) is **not** implemented in this initial version — flagged as an open decision in the spec.
- Respondent-key joins are explicit `left join`s so orphaned rows surface via `not_null`/PK constraints instead of being silently dropped.

## Open questions / follow-ups
- Whether to build delimited multi-select parsing (and how to configure delimiters per integration).
- Participation-rate denominators (survey eligibility) are out of scope until source data is available.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/models/core_warehouse/dim_survey.sql
+++ b/models/core_warehouse/dim_survey.sql
@@ -1,0 +1,29 @@
+{{
+    config(
+        post_hook=[
+            "alter table {{ this }} alter column k_survey set not null",
+            "alter table {{ this }} add primary key (k_survey)",
+        ]
+    )
+}}
+
+with stg_surveys as (
+    select * from {{ ref('stg_ef3__surveys') }}
+),
+-- Intermediate CTE w/ prefixed field references to make adding complexity later easier
+formatted as (
+    select
+        stg_surveys.k_survey,
+        stg_surveys.tenant_code,
+        stg_surveys.survey_id,
+        stg_surveys.survey_title,
+        stg_surveys.survey_category,
+        stg_surveys.school_year,
+        stg_surveys.ed_org_id,
+        stg_surveys.ed_org_type,
+        stg_surveys.school_id,
+        stg_surveys.session_name,
+        stg_surveys.number_administered
+    from stg_surveys
+)
+select * from formatted
--- a/models/core_warehouse/dim_survey_question.sql
+++ b/models/core_warehouse/dim_survey_question.sql
@@ -1,0 +1,25 @@
+-- TODO: make indentation more consistent
+{{
+  config(
+    post_hook=[
+        "alter table {{ this }} alter column k_survey_question set not null",
+        "alter table {{ this }} add primary key (k_survey_question)",
+        "alter table {{ this }} add constraint fk_{{ this.name }}_survey foreign key (k_survey) references {{ ref('dim_survey') }}",
+    ]
+  )
+}}
+
+with stg_survey_questions as (
+    select * from {{ ref('stg_ef3__survey_questions') }}
+),
+formatted as (
+    select
+        stg_survey_questions.k_survey_question,
+        stg_survey_questions.k_survey,
+        stg_survey_questions.tenant_code,
+        stg_survey_questions.question_code,
+        stg_survey_questions.question_text,
+        stg_survey_questions.question_form
+    from stg_survey_questions
+)
+select * from formatted
--- a/models/core_warehouse/dim_survey_question.yml
+++ b/models/core_warehouse/dim_survey_question.yml
@@ -1,0 +1,31 @@
+version: 2
+
+models:
+  - name: dim_survey_question
+    description: "Survey question definitions"
+    config:
+      tags: ['survey', 'core']
+      enabled: "{{ var('src:domain:survey:enabled', True) }}"
+    columns:
+      - name: k_survey_question
+        description: "Surrogate key for survey question"
+        tests:
+          - unique
+          - not_null
+
+      - name: k_survey
+        description: "Foreign key to dim_survey"
+        tests:
+          - not_null
+          - relationships:
+              to: ref('dim_survey')
+              field: k_survey
+
+      - name: question_code
+        description: "Code or identifier for the question"
+
+      - name: question_text
+        description: "Full text of the survey question"
+
+      - name: question_form
+        description: "Form or type of question (e.g., multiple choice, free text)"
--- a/models/core_warehouse/fct_survey_question_response.sql
+++ b/models/core_warehouse/fct_survey_question_response.sql
@@ -1,0 +1,39 @@
+{{
+  config(
+    post_hook=[
+        "alter table {{ this }} alter column k_survey_response set not null",
+        "alter table {{ this }} alter column k_survey_question set not null",
+        "alter table {{ this }} add primary key (k_survey_response, k_survey_question)",
+        "alter table {{ this }} add constraint fk_{{ this.name }}_survey_response foreign key (k_survey_response) references {{ ref('fct_survey_response') }}",
+        "alter table {{ this }} add constraint fk_{{ this.name }}_survey_question foreign key (k_survey_question) references {{ ref('dim_survey_question') }}",
+        "alter table {{ this }} add constraint fk_{{ this.name }}_survey foreign key (k_survey) references {{ ref('dim_survey') }}",
+    ]
+  )
+}}
+
+with stg_survey_question_responses as (
+    select * from {{ ref('stg_ef3__survey_question_responses') }}
+),
+fct_survey_response as (
+    select * from {{ ref('fct_survey_response') }}
+),
+formatted as (
+    select
+        stg_survey_question_responses.k_survey_response,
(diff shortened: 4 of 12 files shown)


models/core_warehouse/dim_survey.sql
```diff
@@ -0,0 +1,29 @@
...
+    )
+}}
+
+with stg_surveys as (
+    select * from {{ ref('stg_ef3__surveys') }}
+),
+-- Intermediate CTE w/ prefixed field references to make adding complexity later easier
+formatted as (
+    select
+        stg_surveys.k_survey,
```
ejoranlienea, line 16: Let's remove `school_id` and `session_name` and replace it with an (optional, nullable) `k_session`.

Session is a weird and kind of misleading inclusion on this model though (by Ed-Fi, I mean) -- there's a lot of ways it could break down, depending how `survey_id` is constructed.

We could also add nullable keys for `k_lea` and `k_school` as we typically do for edorg associations, but given that it's an optional and non-key field, I do wonder about how that could be misused or misinterpreted.

models/core_warehouse/fct_survey_response.sql
```diff
@@ -0,0 +1,37 @@
...
+            when stg_survey_responses.k_staff is not null then 'staff'
+            when stg_survey_responses.k_contact is not null then 'contact'
+            else 'anonymous'
+        end as respondent_type,
+        stg_survey_responses.electronic_mail_address,
+        stg_survey_responses.full_name,
+        stg_survey_responses.location,
+        stg_survey_responses.response_date,
+        stg_survey_responses.completion_time_seconds,
+        1 as response_count
```
ejoranlienea, line 34: Do we need this? Could just `count(*)`, no?

models/core_warehouse/fct_survey_response.sql
```diff
@@ -0,0 +1,37 @@
...
+        stg_survey_responses.k_student,
+        stg_survey_responses.k_staff,
+        stg_survey_responses.k_contact as k_parent,
+        stg_survey_responses.tenant_code,
+        stg_survey_responses.survey_response_id,
+        case
+            when stg_survey_responses.k_student is not null then 'student'
+            when stg_survey_responses.k_staff is not null then 'staff'
+            when stg_survey_responses.k_contact is not null then 'contact'
+            else 'anonymous'
```
rlittle08, line 27: i think we'll need the option to link to a `k_person` as well.. unless we resolve personRefs to one of the above

I also generally wonder if 'anonymous' is correct, when it could be just 'unknown'? maybe if we solve the personRef we can call it anonymous?

models/core_warehouse/fct_survey_response.sql
rlittle08: seconding^

models/core_warehouse/dim_survey.sql
rlittle08: i slight prefer including k_lea and k_school, even though they will sometimes both be null

models/core_warehouse/dim_survey_question.sql
```diff
@@ -0,0 +1,25 @@
...
+    select * from {{ ref('stg_ef3__survey_questions') }}
+),
+formatted as (
+    select
+        stg_survey_questions.k_survey_question,
+        stg_survey_questions.k_survey,
+        stg_survey_questions.tenant_code,
+        stg_survey_questions.question_code,
+        stg_survey_questions.question_text,
+        stg_survey_questions.question_form
```
rlittle08, line 22: i think we should include v_response_choices, v_matrices, and k_survey_section here

models/core_warehouse/fct_survey_question_response.sql
```diff
@@ -0,0 +1,39 @@
...
+        fct_survey_response.k_student,
+        fct_survey_response.k_staff,
+        fct_survey_response.k_parent,
+        stg_survey_question_responses.tenant_code,
+        fct_survey_response.respondent_type,
+        fct_survey_response.location,
+        stg_survey_question_responses.comment,
+        stg_survey_question_responses.no_response,
+        fct_survey_response.response_date,
+        1 as question_response_count
```
rlittle08, line 34: same comment as above

models/core_warehouse/fct_survey_question_response.sql
```diff
@@ -0,0 +1,39 @@
...
+    from stg_survey_question_responses
+    left join fct_survey_response
```
rlittle08, line 36: shouldn't this be inner?

rlittle08, line 36: also, might as well inner join to dim_survey_question, we usually will do that as safeguard against surfacing faulty records

models/core_warehouse/fct_survey_question_response.sql
```diff
@@ -0,0 +1,39 @@
...
+}}
+
+with stg_survey_question_responses as (
+    select * from {{ ref('stg_ef3__survey_question_responses') }}
+),
+fct_survey_response as (
+    select * from {{ ref('fct_survey_response') }}
+),
+formatted as (
+    select
```
rlittle08, line 21: similar to my commen in fct_survey_response, i think we should include the full variant v_survey_question_response_values here..

but that does make me wonder if we should actually remove this fct entirely, if 99% of analysis will be routed to fct_survey_question_response_value anyway?

models/core_warehouse/dim_survey_question.yml
```diff
@@ -0,0 +1,31 @@
...
+        description: "Surrogate key for survey question"
+        tests:
+          - unique
+          - not_null
+
+      - name: k_survey
+        description: "Foreign key to dim_survey"
+        tests:
+          - not_null
+          - relationships:
```
rlittle08, line 20: please also add `constraints:` with the foreign keys to all models (this is recent required practice for edu_wh models)

models/core_warehouse/fct_survey_question_response_value.sql
```diff
@@ -0,0 +1,43 @@
...
+        fct_survey_question_response.k_parent,
+        fct_survey_question_response.respondent_type,
+        stg_survey_question_response_values.numeric_response,
+        stg_survey_question_response_values.text_response,
+        fct_survey_question_response.location,
+        fct_survey_question_response.comment,
+        fct_survey_question_response.no_response,
+        fct_survey_question_response.response_date,
+        1 as question_response_value_count
+    from stg_survey_question_response_values
```
rlittle08, line 38: should add inner joins to dims for extra safety on ref integrity

models/core_warehouse/fct_survey_question_response_value.sql
```diff
@@ -0,0 +1,43 @@
...
+        stg_survey_question_response_values.k_survey_response,
+        stg_survey_question_response_values.k_survey_question,
+        stg_survey_question_response_values.tenant_code,
+        stg_survey_question_response_values.question_response_value_id,
+        fct_survey_question_response.k_survey,
+        fct_survey_question_response.k_student,
+        fct_survey_question_response.k_staff,
+        fct_survey_question_response.k_parent,
+        fct_survey_question_response.respondent_type,
+        stg_survey_question_response_values.numeric_response,
```
rlittle08, line 31: i think it's very likely that we need xwalks for these responses in order to map them to analytical meaning, e.g. order, standardized values, etc. - i assume those are needed for the podium dashboards, so can we go ahead and add now?

rlittle08 (changes requested):
overall summary:
- this is a clean, solid foundation. looks on track for end of month release
- please add foreign key constraints
- I want to discuss utility of `fct_survey_question_response` - what analysis does this enable, that couldn't be done from `fct_survey_question_response_value`?

PR #248, opened 2026-06-03
Fix/active enrollment flag enhancement

smckee-tnedu:
## Description & motivation
Tennessee Districts start loading data for the next school year into a new ODS in the March/April timeframe. They are doing this to get their Calendars ready. Stadium_Tennessee has a suite of business rules which will alert them when their submitted Calendars don't follow TNDOE rules. We use our the front end to EDU that we've coded to show them these errors.
BUT BECAUSE they start loading in data in the March/April timeframe, the current code for fct_student_school_association starts setting active_enrollment to false. There is a portion of the code that checks to see what the latest school year of loaded data is and, for us, since it's future, things get set to false.
I tried to make this fit in with the new "buffer_days" code for 0.6.3, but I admit that I don't totally get the reasoning behind the implementation. So I'm not positive that what I've done doesn't somehow break the intention of that impl.

Also, I'm sorry if this is showing as a merge conflict because of the way I did it. We are on 0.6.2 and I made these changes against what exists in main by just copying the code in your main and applying my logic overtop. I probably could (should?) have cherry picked the proper commit from your branch so that there was no conflict but I'm short on time. You'll have to manually diff your main branch and this change to everything I changed properly. Maybe. I don't know.

## Breaking changes introduced by this PR:
The active_enrollment flag might not be set to what people are expecting. But I have no idea. It is working for us, but we don't use buffer_days.

## PR Merge Priority:
- [X] Low
- [ ] Medium
- [ ] High

## Changes to existing files:
- `fct_student_school_association` : Modified the logic for is_active_enrollment so that the future school year data doesn't matter. Whether or not an enrollment is active is entirely determined by if the current_date() is both within the enrollment window and the school year window. I didn't change the buffer_days logic, but if I understand it correctly, it should still work as intended.

## New files created:
None

## Tests and QC done:
I have done unit tests on my side and this is working as expected. It is about to go through our QA process though. It has not yet.

## edu_wh PR Review Checklist:
Make sure the following have been completed before approving this PR:
- [ ] Description of changes has been added to Unreleased section of [CHANGELOG.md](/CHANGELOG.md). Add under `## New Features` for features, etc.
- [ ] Code has been tested/checked for Databricks and Snowflake compatibility - EA engineers see Databricks checklist [here](https://edanalytics.slite.com/app/docs/LRjXFxVRAWA5ST) 
- [ ] Reviewer confirms the grain of all tables are unchanged, OR any changes are expected, communicated, and this PR is flagged as a breaking change (not for patch release)
[...]

PR #270, opened 2026-08-23
Refactor/fct_student_school_association active enrollment flag

AngelicaLastra:
Related to TN PR #248 

# Description & motivation

This PR moves enrollment validation and active-enrollment logic out of `fct_student_school_association` and into a new build model, `bld_ef3__stu_sch_assoc__enrollment_flags`.

Before:
enrollment__before

After:
enrollment__after

## Main Problem Tackled
The main behavior change is a fix for tenants that load next year's enrollment data early. Before, as soon as a future school year appeared in the data, that year became the "active" school year. Current-year enrollments would immediately flip to inactive, even though the future school year had not started yet.

active_enrollment_errors

## Refactoring Goals
The fact table keeps the same output columns as before, so downstream models do not need to change.

The refactor also untangles a lot of logic that had all been living in `fct_student_school_association`. Before this change, that one fact model was responsible for all of the following:

- validating enrollment records.
- resolving school calendars.
- calculating `is_active_enrollment` inside one dense `iff()`.
- outputting the final fact-table columns.

This meant you had to read through a lot of different concerns at once just to understand how active enrollment was being decided. It also made the logic harder to reuse elsewhere and made bugs like the active-school-year issue harder to debug externally.

With this PR, those responsibilities are divided into two models:

- `bld_ef3__stu_sch_assoc__enrollment_flags` owns enrollment validation, calendar resolution, and the component flags that describe enrollment status.
- `fct_student_school_association` combines those flags into `is_active_enrollment` and outputs the final schema/fields.

## Optional Walkthrough of new logic with Examples

<details>
  <summary> Click to expand </summary>

---
Let's say we have a district, District A, and it is currently mid-July 2026. 
The district decides to load records for the upcoming 2027 school year early in order to prep for that year, which begins August 25, 2026.

`stg_ef3__student_school_associations` contains these four records:

| row | school_year | entry_date | exit_withdraw_date | exit_withdraw_type |
| --- | --- | --- | --- | --- |
| Bob | 2026 | 2025-08-20 | null | null |
| Sandy | 2026 | 2025-08-20 | 2025-08-10 | Withdrawn |
| Pearl | 2026 | 2025-09-01 | 2025-09-01 | No show |
| Patrick | 2027 | 2026-08-25 | null | null |

These represent four different situations:

- **Bob** is a student who enrolled last August and is still enrolled.
- **Sandy** is invalid because the exit date is before the entry date.
- **Pearl** is a no-show who was recorded but never actually attended.
- **Patrick** is a valid pre-registration for the next school year.

### Step 1: resolve enrollment configuration once

The build model starts by resolving every `edu:enroll:*` var it uses.

This includes configuration such as:

- which withdrawal codes should remove an enrollment.
- whether exit dates are inclusive.
- whether enrollment-date filters are enabled.
- how the year-end active-enrollment buffer behaves.

Implementations can still override these values in their own `dbt_project.yml`.

The goal is to have these configurations be more readable, so instead of finding `var()` calls and `{% set %}` statements scattered throughout the SQL, you can see all of the enrollment configuration in one place before getting into the actual logic.

### Step 2: join the related dimensions and name each validity check

The `stu_school_records` CTE joins staging to `dim_student` and `dim_school`.

These are inner joins, so records that cannot be tied to a valid student or school drop here.

The CTE also resolves `k_school_calendar`:

- use the directly matched calendar when one exists.
- otherwise use the school's only calendar when there is exactly one clear option.

It then calculates four enrollment validity flags, each are phrased positively so `true` means the enrollment should survive that rule.

For the example records, some of those checks look like this:

| row | school_year | entry_date | exit_withdraw_date | exit_withdraw_type |is_exit_on_or_after_entry | has_retained_withdraw_type | is_exit_on_or_after_first_day |
| --- | --- | --- | --- | --- |--- | --- | --- |
| Bob | 2026 | 2025-08-20 | null | null |true | true | true |
| Sandy | 2026 | 2025-08-20 | 2025-08-10 | Withdrawn |**false** | true | true |
| Pearl | 2026 | 2025-09-01 | 2025-09-01 | No show |true | **false** | true |
| Patrick | 2027 | 2026-08-25 | null | null |true | true | true |

Before, these same checks were all folded into one larger `where` clause. Now each one is separated into its own field so it is much easier to see what is being checked and why a record would be kept or removed.

### Step 3: filter invalid records before calculating enrollment status

`valid_stu_school_records` applies the filters based on validity flags set above.

In the example:

- Sandy is removed because its exit date is before its entry date.
- Pearl is removed because its withdrawal type is configured to exclude the record.
- Bob and Patrick remain.

After filtering, the helper validity flags are dropped with `star(except=...)` because they won't need to be used downstream, so avoiding the noise is valuable.

The next step decides which school year is active using a `max()` window.

### Step 4: calculate the component enrollment-status flags

`enrollment_status_flags` adds the three pieces that eventually make up `is_active_enrollment`:

- `is_active_school_year`.
- `has_enrollment_started`.
- `is_within_effective_end_date`.

Breaking these into separate flags makes the final `is_active_enrollment` definition much easier to read, and it also means another model can reuse one of those pieces without having to recreate the full active-enrollment calculation.

#### `is_active_school_year`

This flag answers:

> Is this enrollment part of the most recent school year that has actually started for this tenant?

This is also where the bug with future school-year data is fixed. 

Before, the logic effectively chose the newest school year present in the data, which meant that in our example, loading Patrick in July caused 2027 to become the active year immediately, even though the 2027 school year does not begin until August 25.

That produced this situation:

| row | school_year | entry_date | exit_withdraw_date | max(school_year) | notes |
| --- | --- | --- | --- | --- |--- |
| Bob | 2026 | 2025-08-20 | null | **false** | this means we are setting this record as inactive even though we are in this active school year
| Patrick | 2027 | 2026-08-25 | null | true |

The problem gets worse because Patrick also has a future `entry_date`.

So during the summer:

- 2026 was no longer considered the effective school year.
- 2027 students had not started yet.
- the tenant ends up showing zero active enrollments until 2027 starts.

The new logic only allows a school year to become active once its first school day has passed.

| row | school_year | entry_date | exit_withdraw_date | max(school_year if the enrollment window has started) | notes |
| --- | --- | --- | --- | --- |--- |
| Bob | 2026 | 2025-08-20 | null | true | now this row remains active.
| Patrick | 2027 | 2026-08-25 | null | **false** |

#### `has_enrollment_started`

This flag answers:

> Has the student's `entry_date` arrived yet?

For the example:

| row | school_year | entry_date | exit_withdraw_date | is_active_school_year | has_enrollment_started | notes |
| --- | --- | --- | --- | --- |--- |--- |
| Bob | 2026 | 2025-08-20 | null | true | true |
| Patrick | 2027 | 2026-08-25 | null | **false** |  **false** | this remains false until 2026-08-25

#### `is_within_effective_end_date`

This flag answers:

> Has the enrollment reached the date when it should stop being considered active?

Normally, this is based on the enrollment's exit date, using the configured inclusive or exclusive behavior.

When the year-end buffer is configured, there is one additional case.

Students who exit within the configured number of days around the final school day can remain active through the end of that calendar year. This prevents summer reporting from immediately losing students the day after the school year ends.

### Step 5: compose the flags in the fact model

After this refactor, `fct_student_school_association` mostly handles presentation/final definition through an `iff()` statement.

It selects from the new build model and calculates:

`is_active_enrollment = is_active_school_year and has_enrollment_started and is_within_effective_end_date`

The result is coalesced to `false` so a null value does not produce a null active-enrollment flag.

The fact still handles the pieces that belong specifically to the final output table:

- joins `xwalk_grade_levels` to calculate the grade integer.
- calculates `is_latest_annual_entry`.
- includes configured extension columns.
- applies the existing post-hook constraints.
- outputs the same explicit column list as before.

For the walkthrough:

- Bob is outputted with `is_active_enrollment = true`.
- Patrick is outputted with `is_active_enrollment = false` until the new school year begins.
- Sandy and Pearl are filtered out.

That is the same expected record behavior as before the refactor, except the summer active-school-year bug is removed.

</details>

## Why split the models this way?

The main point of separating the logic is to make it clearer, more accessible, and easier to maintain/extend. 

In this PR, the build model owns the creation of the flags:

- record validation.
- calendar resolution.
- active-school-year selection.
- enrollment-start status.
- effective-end-date status.

The fact owns the final definition that downstream apps/tables can use:

- what combination of those flags means `is_active_enrollment`.
- what columns are outputted in the final version.

It also gives us a chance to extend what flags we include as active enrollment in the future, and make this configurable if the need arises.

# PR Merge Priority

- [x] Low
- [ ] Medium
- [ ] High

# Changes to existing files

## `fct_student_school_association`

Now selects from `bld_ef3__stu_sch_assoc__enrollment_flags` instead of staging.

Changes include:

- Removed the enrollment-validity `where` logic.
- Removed the school-calendar resolution joins.
- Removed the active-school-year window logic.
- Moved those responsibilities into the new build model.
- Changed `is_active_enrollment` to combine the build model's three status flags and coalesce the result to `false`.
- Removed an unused `dim_calendar_date` CTE that was declared but never joined or selected.

The following behavior remains in the fact:

- `xwalk_grade_levels` join.
- `is_latest_annual_entry`.
- extension column roll call through `extract_extension` with `flatten=False`.
- existing post-hook constraints.
- the same final output column set.

## `CHANGELOG.md`

Added two Unreleased entries:

- an "Under the hood" entry for the model refactor.
- a "Fixes" entry for the active-school-year behavior change.

## `models/build/edfi_3/_edfi_3.yml`

Registered the new build model with the `core` tag and kept the model list alphabetical.

# New files created

## `bld_ef3__stu_sch_assoc__enrollment_flags`

New build model at the same grain as the staged enrollment record after deduplication:

`k_student, k_school, entry_date`

The model produces validated enrollment records with a resolved `k_school_calendar` and three enrollment-status flags:

- `is_active_school_year`.
- `has_enrollment_started`.
- `is_within_effective_end_date`.

The model is broken into four main pieces:

1. A configuration block that resolves all `edu:enroll:*` vars used by the model.
2. A join CTE that resolves related dimensions and calculates the named validity flags.
3. A filter CTE that applies the validity rules and removes the helper flags.
4. A final CTE that calculates the three enrollment-status flags.

# Behavior change to be aware of

No new vars are introduced, and the meanings of the existing `edu:enroll:*` vars do not change.

However, `is_active_enrollment` can change for tenants that load next year's enrollment data before that school year begins.

Before, loading the future year caused the current year to stop being considered active immediately. Because the future enrollments had not reached their entry dates yet, this could leave the tenant with zero active enrollments until the new school year started.

Now, the current school year remains active until the future school year's first school day has actually passed.

# Tests and QC done

- [x] Compared the build model's filtering, calendar resolution, and status logic clause by clause against the original fact logic, including year-end buffer sign handling and exit-date inclusivity fallbacks.
- [x] Run `dbt build -s +fct_student_school_association`
- [x] Compared outputs and `is_active_enrollment` values against the pre-refactor table to confirm the refactor is functioning properly and outputting the same rows as before

### DIFF
--- a/models/build/edfi_3/bld_ef3__stu_sch_assoc__enrollment_flags.sql
+++ b/models/build/edfi_3/bld_ef3__stu_sch_assoc__enrollment_flags.sql
@@ -1,0 +1,163 @@
+with stg_stu_school as (
+    select * from {{ ref('stg_ef3__student_school_associations') }}
+),
+
+dim_student as (
+    select * from {{ ref('dim_student') }}
+),
+
+dim_school as (
+    select * from {{ ref('dim_school') }}
+),
+
+dim_school_calendar as (
+    select * from {{ ref('dim_school_calendar') }}
+),
+
+bld_school_calendar_windows as (
+    select * from {{ ref('bld_ef3__school_calendar_windows') }}
+),
+
+single_calendar_schools as (
+    -- some implementations may not provide a school calendar link because
+    -- their system only allows one calendar per school anyway.
+    -- we will detect these cases and fill in missing school calendars
+    -- only when doing so is unambiguous
+    select * from dim_school_calendar
+    qualify 1 = count(*) over(partition by k_school, school_year)
+),
+
+valid_stu_school_records as (
+
+    select
+        -- exclude the raw calendar reference so the resolved value below
+        -- can take over the k_school_calendar name
+        {{ edu_edfi_source.star('stg_stu_school', except=['k_school_calendar']) }},
+
+        dim_school.k_lea,
+
+        -- the enrollment's raw calendar reference, resolved: the direct match
+        -- against dim_school_calendar when it exists, otherwise the school's
+        -- only calendar when that choice is unambiguous
+        coalesce(
+            dim_school_calendar.k_school_calendar,
+            single_calendar_schools.k_school_calendar
+        ) as k_school_calendar,
+
+        -- first_school_day to decide whether a school year has begun,
+        -- last_school_day for the year-end buffer check
+        bld_school_calendar_windows.first_school_day,
+        bld_school_calendar_windows.last_school_day
+
+    from stg_stu_school
+    -- these joins are filters as much as lookups: an enrollment with no
+    -- matching student or school row is dropped, and must be dropped here so
+    -- it does not vote in the is_active_school_year window below
+    join dim_student
+        on stg_stu_school.k_student = dim_student.k_student
+    join dim_school
+        on stg_stu_school.k_school = dim_school.k_school
+    left join dim_school_calendar
+        on stg_stu_school.k_school_calendar = dim_school_calendar.k_school_calendar
+    left join single_calendar_schools
+        on stg_stu_school.k_school = single_calendar_schools.k_school
+        and stg_stu_school.school_year = single_calendar_schools.school_year
+    -- the windows join keys on the direct calendar match only, never the
+    -- single-calendar fallback
+    left join bld_school_calendar_windows
+        on stg_stu_school.k_school = bld_school_calendar_windows.k_school
+        and stg_stu_school.school_year = bld_school_calendar_windows.school_year
+        and equal_null(
+            dim_school_calendar.k_school_calendar,
+            bld_school_calendar_windows.k_school_calendar
+        )
+
+    where true
+    {%- if var('edu:enroll:exclude_exit_before_first_day', True) %}
+      {#- allow implementations with exclusive exit dates to still keep first-day exits -#}
+      {%- set first_day_inclusive = var('edu:enroll:first_day_exit_date_inclusive', var('edu:enroll:exit_withdraw_date_inclusive', True)) %}
+        -- exclude students who exited before the first school day
+        and (
+            {{ date_within_end_date(
+                'bld_school_calendar_windows.first_school_day',
+                'stg_stu_school.exit_withdraw_date',
+                first_day_inclusive
+            ) }}
+            or bld_school_calendar_windows.first_school_day is null
+        )
+    {%- endif %}
+        -- make sure exit occurred on or after entry
+        and (
+            stg_stu_school.exit_withdraw_date is null
+            or stg_stu_school.exit_withdraw_date >= stg_stu_school.entry_date
+        )
+    {%- set excl_withdraw_codes = var('edu:enroll:exclude_withdraw_codes') %}
+    {%- if excl_withdraw_codes is string %}
+      {%- set excl_withdraw_codes = [excl_withdraw_codes] %}
+    {%- endif %}
+    {%- if excl_withdraw_codes | length %}
+        -- drop invalid enrollments while retaining rows without an exit type
+        and (
+            stg_stu_school.exit_withdraw_type not in (
+                '{{ excl_withdraw_codes | join("', '") }}'
+            )
+            or stg_stu_school.exit_withdraw_type is null
+        )
+    {%- endif %}
+    {%- if var('edu:enroll:exclude_cross_year_enrollments', False) %}
+        -- drop enrollments whose year disagrees with the student record's year
+        and dim_student.school_year = stg_stu_school.school_year
+    {%- endif %}
+
+),
+
+enrollment_status_flags as (
+
+    select
+        *,
+
+        -- is this row's school year the newest one that has actually begun
+        -- for this tenant? falls back to entry_date when the calendar
+        -- didn't match, so tenants with missing calendar data aren't
+        -- stranded. using the newest year *loaded* would strand current
+        -- enrollments whenever next year's data is ingested early.
+        school_year = max(
+            iff(
+                coalesce(first_school_day, entry_date) <= current_date(),
+                school_year,
+                null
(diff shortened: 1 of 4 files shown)


models/build/edfi_3/bld_ef3__stu_sch_assoc__enrollment_flags.sql
```diff
@@ -0,0 +1,243 @@
...
+
+    select
+        *,
+
+        -- identify the school year that should currently count as active for each tenant.
+        -- only consider school years whose first school day has already passed.
+        -- example: 
+            -- if 2027 enrollment records are loaded in july 2026,
+            -- but the 2027 school year starts in august, 2026 remains the active school year until then.
+        school_year = max(
```
rlittle08, line 205: this logit makes a lot of sense, an is cleanly written 👍

rlittle08 (approved):
this looks great, big improvement, and i tested to complete success in TX. LGTM

smckee-tnedu:
This looks great! I can't wait to upgrade to the release with this change to see how it works in TN.
Someday... when they give me time to do so. lol

